PR #6, opened 2020-04-18
add install instructions

edgarriba:
# Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?

## What does this PR do?
Fixes # (issue).

## PR review    

## Did you have fun?


requirements.txt
```diff
@@ -1,3 +1,4 @@
 torch>=1.1
 pytorch-lightning>=0.7.1
-torchvision>=0.4.0
\ No newline at end of file
+torchvision>=0.4.0
```
williamFalcon, line 3: Can we add these to the requirements-extra?
@Borda or does it make sense to allow these into reqs here?

i can imagine bolts getting bloated if we don't keep this slim? especially as we add submodules?

edgarriba: which submodules will depend on vision stuff and which not ?

Borda: I am thinking that we set as only requirements pytorch-lightning in setup file and all in requirements.txt are in fact extras...

Borda: Altertive may be requirements.txt per submodules and in root link them together

williamFalcon: but you auto read requirements.txt, so that won't work.

williamFalcon: not every model here will be vision based (ie: audio, text, etc...). and we also have callbacks and such.
If someone only wants callbacks they don't need to install torchvision or anything image related.

edgarriba: my guess is that you will have to have predefined requirements lists for each subggroup

justusschock: yep, that way you could do something like `pip install bolts[vision]`

Borda:
@edgarriba just rebased your PR, mind check that is it still correct with respect to large Bolts changes...
cc: @williamFalcon

PR #26, opened 2020-05-18
SSL plumbing

williamFalcon:
AMDIM has verified SOTA on imagenet.
Running CPC now


Borda:
come issue with
```
class CIFAR10Mixed(SSLDatasetMixin, CIFAR10):
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

PR #95, opened 2020-06-30
Dlrm

williamFalcon:
breaks docs for some reason


williamFalcon:
@Borda for some reason this breaks the docs...

Borda:
> Borda for some reason this breaks the docs...

what does it break, do you mean the Conv section?

PR #118, opened 2020-07-15
Confused logit callback

williamFalcon:
Takes the logit predictions of a model and when the probabilities of two classes are very close, the model
        doesn't have high certainty that it should pick one vs the other class.

        This callback shows how the input would have to change to swing the model from one label prediction
        to the other.


pl_bolts/callbacks/vision/confused_logit.py
```diff
@@ -0,0 +1,129 @@
...
+
+        """
+        super().__init__()
+        self.top_k = top_k
+        self.projection_factor = projection_factor
+
+    def on_batch_end(self, trainer, pl_module):
+
+        # show images only every 20 batches
+        if (trainer.batch_idx + 1) % 20 != 0:
```
Borda, line 48: this 20 shall be init argument

pl_bolts/callbacks/vision/confused_logit.py
```diff
@@ -0,0 +1,129 @@
...
+            return
+
+        # pick the last batch and logits
+        # TODO: use context instead
+        x, y = pl_module.last_batch
+        l = pl_module.last_logits
+
+        # only check when it has opinions (ie: the logit > 5)
+        if l.max() > 5.0:
```
Borda, line 57: this confidence could be also class parameter

pl_bolts/callbacks/vision/confused_logit.py
```diff
@@ -0,0 +1,129 @@
...
+
+        for img_i in range(len(confusing_x)):
+            x = confusing_x[img_i].squeeze(0)
+            y = confusing_y[img_i]
+            ga = grad_a[img_i]
+            gb = grad_b[img_i]
+
+            mask_idx = mask_idxs[img_i]
+
+            fig = plt.figure(figsize=(15, 10))
```
Borda, line 103: use rather `fig, axarr = plt.subplots(ncol=3, nrow=2)`
and alter you just `axarr[1, 2].imshow(...)`

williamFalcon: i don't have a lot of time to mess with this. i don't know if this will work, but i know the current version works.
If you feel strongly about it, mind making the changes and posting a colab that shows this works?

Borda: ok I ll check I because the call to `plt.` takes the last pointer and if you plot two in parallel you will write everything into one

PR #133, opened 2020-08-02
Loaders

williamFalcon:
@nateraw... can't have args in dataloaders since PL calls them without args.

### DIFF
--- a/docs/source/datamodules.rst
+++ b/docs/source/datamodules.rst
@@ -102,13 +102,13 @@ Example::
         def prepare_data(self):
             # download and do something to your data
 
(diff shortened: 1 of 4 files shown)


williamFalcon:
@nateraw... for datamodules, batch_size should be an arg to init.

docs/source/datamodules.rst
```diff
@@ -102,13 +102,13 @@ Example::
         def prepare_data(self):
             # download and do something to your data
 
-        def train_dataloader(self, batch_size):
+        def train_dataloader(self):
```
Borda, line 105: this is the reason why all these blocked shall be tested...

PR #138, opened 2020-08-03
Linear warmup sched

ananyahjha93:
# Before submitting

- [x] Did you make sure to update the docs?   

## What does this PR do?
Fixes # (issue).

## PR review    

## Did you have fun?

### DIFF
--- a/pl_bolts/optimizers/lr_scheduler.py
+++ b/pl_bolts/optimizers/lr_scheduler.py
@@ -2,12 +2,11 @@ import math
 import warnings
 
 from torch.optim.lr_scheduler import _LRScheduler
-from torch.optim.lr_scheduler import ReduceLROnPlateau
-from torch.optim.lr_scheduler import LambdaLR
 
 
+"""
 class GradualWarmupScheduler(_LRScheduler):
-    """ Gradually warm-up(increasing) learning rate in optimizer.
+    "" Gradually warm-up(increasing) learning rate in optimizer.
     Proposed in 'Accurate, Large Minibatch SGD: Training ImageNet in 1 Hour'.
 
     Adapted from: https://github.com/ildoonet/pytorch-gradual-warmup-lr
@@ -17,7 +16,7 @@ class GradualWarmupScheduler(_LRScheduler):
         multiplier: target learning rate = base lr * multiplier if multiplier > 1.0. if multiplier = 1.0, lr starts from 0 and ends up with the base_lr.
         total_epoch: target learning rate is reached at total_epoch, gradually
         after_scheduler: after target_epoch, use this scheduler(eg. ReduceLROnPlateau)
-    """
+    ""
 
     def __init__(self, optimizer, multiplier, total_epoch, after_scheduler=None):
         self.multiplier = multiplier
@@ -70,6 +69,7 @@ class GradualWarmupScheduler(_LRScheduler):
                 return super(GradualWarmupScheduler, self).step(epoch)
         else:
             self.step_ReduceLROnPlateau(metrics, epoch)
+"""
 
 
 class LinearWarmupCosineAnnealingLR(_LRScheduler):
--- a/pl_bolts/optimizers/lr_scheduler.py
+++ b/pl_bolts/optimizers/lr_scheduler.py
@@ -1,0 +1,141 @@
+import math
+import warnings
+
+from torch.optim.lr_scheduler import _LRScheduler
+from torch.optim.lr_scheduler import ReduceLROnPlateau
+from torch.optim.lr_scheduler import LambdaLR
+
+
+class GradualWarmupScheduler(_LRScheduler):
+    """ Gradually warm-up(increasing) learning rate in optimizer.
+    Proposed in 'Accurate, Large Minibatch SGD: Training ImageNet in 1 Hour'.
+
+    Adapted from: https://github.com/ildoonet/pytorch-gradual-warmup-lr
+
+    Args:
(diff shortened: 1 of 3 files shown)


pl_bolts/optimizers/lr_scheduler.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if self.last_epoch == 0:
+            return [self.warmup_start_lr] * len(self.base_lrs)
+        elif self.last_epoch < self.warmup_epochs:
+            return [group['lr'] + (base_lr - self.warmup_start_lr) / (self.warmup_epochs - 1)
+                    for base_lr, group in zip(self.base_lrs, self.optimizer.param_groups)]
+        elif self.last_epoch == self.warmup_epochs:
+            return self.base_lrs
+        # TODO: what does this do? This might be wrong for now since axis has shifted due to warmup
+        elif (self.last_epoch - 1 - self.max_epochs) % (2 * (self.max_epochs - self.warmup_epochs)) == 0:
```
ananyahjha93, line 62: @williamFalcon @Borda this is the T_curr = 2 (k + 1) T_max case for linear warmup + cosine annealing scheduler. This should fail but it doesn't. Only when you adjust it for the linear warmup part it fails. This elif condition is not adjusted for the linear warmup case.

tests/unit/optimizers/test_lr_scheduler.py
```diff
@@ -0,0 +1,367 @@
...
+                assert abs(target[epoch] - param_group['lr']) < EPSILON,\
+                    "LR is wrong in epoch {}: expected {}, got {}".format(
+                        epoch, target[epoch], param_group['lr']
+                    )
+            [scheduler.step() for scheduler in schedulers]
+
+    def _test_against_closed_form(self, scheduler, closed_form_scheduler, epochs=10):
+        targets = []
+        for epoch in range(epochs):
+            closed_form_scheduler.step(epoch)
```
ananyahjha93, line 65: this comes from earlier versions of pytorch where scheduler.step() was called before the epoch instead of after.

ananyahjha93:
@Borda feedback on formatting as well, used https://github.com/psf/black on @nateraw 's advice.

pl_bolts/optimizers/lr_scheduler.py
```diff
@@ -0,0 +1,144 @@
...
+        ...     scheduler.step()
+        >>> #
+        >>> # passing epoch param case
+        >>> for epoch in range(40):
+        ...     scheduler.step(epoch)
+        ...     # train(...)
+        ...     # validate(...)
+    """
+
+    def __init__(
```
justusschock, line 53: can we add type annotations & docstrings here?

pl_bolts/optimizers/lr_scheduler.py
```diff
@@ -0,0 +1,144 @@
...
+    ):
+
+        self.warmup_epochs = warmup_epochs
+        self.max_epochs = max_epochs
+        self.warmup_start_lr = warmup_start_lr
+        self.eta_min = eta_min
+
+        super(LinearWarmupCosineAnnealingLR, self).__init__(optimizer, last_epoch)
+
+    def get_lr(self):
```
justusschock, line 70: Also return type annotations & docstrings :)

pl_bolts/optimizers/lr_scheduler.py
```diff
@@ -0,0 +1,144 @@
...
+                    * (self.last_epoch - self.warmup_epochs - 1)
+                    / (self.max_epochs - self.warmup_epochs)
+                )
+            )
+            * (group["lr"] - self.eta_min)
+            + self.eta_min
+            for group in self.optimizer.param_groups
+        ]
+
+    def _get_closed_form_lr(self):
```
justusschock, line 121: Type annotations & docstrings as well :)

tests/unit/optimizers/test_lr_scheduler.py
```diff
@@ -0,0 +1,434 @@
...
+from torch.optim.lr_scheduler import _LRScheduler
+
+import math
+import numpy as np
+from tests import reset_seed
+import pytorch_lightning as pl
+from pl_bolts.optimizers.lr_scheduler import LinearWarmupCosineAnnealingLR
+
+
+EPSILON = 1e-12
```
justusschock, line 13: I think there is an epsilon defined somewhere in lightning iirc, maybe use this to avoid duplication?

cc @Borda who might know, where it is ^^

tests/unit/optimizers/test_lr_scheduler.py
justusschock: but do we still support these old versions?

ananyahjha93: @justusschock current version of pytorch still defines ``` _test_against_closed_form()``` for now, all we need to do is define that function to support the old version. I have a warning and doc available for this

tests/unit/optimizers/test_lr_scheduler.py
ananyahjha93: @Borda so I had this question about default epsilons in lightning, ```numpy.finfo(numpy.float32).eps``` is 1.1920929e-07. Since learning rates used are up to 1e-6, we need a higher precision in schedulers. But, ```numpy.finfo(numpy.float64).eps``` comes out to be 2.220446049250313e-16, which results in failing tests. So the ideal epsilon seemed 1e-12.

PR #157, opened 2020-08-14
Faster RCNN Model + Pascal VOC DataModule

teddykoker:
# Before submitting

- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you make sure to update the docs?   
- [X] Did you write any new necessary tests?  

## What does this PR do?

1. Implements the Pascal VOC Detection dataset as a `DataModule`
2. Implements Faster RCNN as a `LightningModule`

## PR review    

## Did you have fun?

### DIFF
--- a/pl_bolts/datamodules/vocdetection_datamodule.py
+++ b/pl_bolts/datamodules/vocdetection_datamodule.py
@@ -2,7 +2,85 @@ import torch
 from pytorch_lightning import LightningDataModule
 from torch.utils.data import DataLoader
 from torchvision.datasets import VOCDetection
-from torchvision import transforms as transform_lib
+import torchvision.transforms as T
+
+
+class Compose(object):
+    def __init__(self, transforms):
+        self.transforms = transforms
+
+    def __call__(self, image, target):
+        for t in self.transforms:
+            image, target = t(image, target)
+        return image, target
+
+
+CLASSES = (
+    "__background__ ",
+    "aeroplane",
+    "bicycle",
+    "bird",
+    "boat",
+    "bottle",
+    "bus",
+    "car",
+    "cat",
+    "chair",
+    "cow",
+    "diningtable",
+    "dog",
+    "horse",
+    "motorbike",
+    "person",
+    "pottedplant",
+    "sheep",
+    "sofa",
+    "train",
+    "tvmonitor",
+)
+
+
+def _prepare_voc_instance(image, target):
+    """
+    Prepares VOC dataset into appropriate target for fasterrcnn
+
+    https://github.com/pytorch/vision/issues/1097#issuecomment-508917489
+    """
+    anno = target["annotation"]
+    h, w = anno["size"]["height"], anno["size"]["width"]
+    boxes = []
+    classes = []
+    area = []
+    iscrowd = []
+    objects = anno["object"]
+    if not isinstance(objects, list):
+        objects = [objects]
+    for obj in objects:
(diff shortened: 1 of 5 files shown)


pl_bolts/datamodules/vocdetection_datamodule.py
```diff
@@ -0,0 +1,199 @@
+import torch
```
Borda, line 1: no need to repeat `_datamodule.py` in the name

pl_bolts/models/detection/faster_rcnn.py
```diff
@@ -0,0 +1,146 @@
...
+        parser.add_argument("--pretrained_backbone", type=bool, default=True)
+        parser.add_argument("--trainable_backbone_layers", type=int, default=3)
+        parser.add_argument("--replace_head", type=bool, default=True)
+
+        parser.add_argument("--data_dir", type=str, default=".")
+        parser.add_argument("--batch_size", type=int, default=1)
+        return parser
+
+
+def cli_main():
```
Borda, line 129:
```suggestion
def run_cli():
```
let's add test for the script too

pl_bolts/datamodules/vocdetection_datamodule.py
teddykoker: I figured I would name it to be consistent with `imagenet_datamodule.py`, `mnist_datamodule.py`, etc. We may just want to rename all of these in the future.

pl_bolts/models/detection/faster_rcnn.py
teddykoker: Question about tests, in this case the cli needs to download the Pascal VOC data, which is 2GB and takes some time, should I still implement?

teddykoker:
@Borda didn't add test for cli, but added test training with `DummyDetectionDataset`, which I just added.

pl_bolts/datamodules/vocdetection_datamodule.py
williamFalcon: i agree with consistency and yes, we need to keep the _datamodule.py name

tests/models/test_detection.py
```diff
@@ -0,0 +1,29 @@
...
+
+    image = torch.rand(1, 3, 400, 400)
+    model(image)
+
+
+def test_fasterrcnn_train(tmpdir):
+
+    model = FasterRCNN()
+
+    train_dl = DataLoader(DummyDetectionDataset(), collate_fn=_collate_fn)
```
williamFalcon, line 25: We have a dummy dataset already. why is this one different?

teddykoker: Instead of a tensor label it returns a dict with values `{"boxes": FloatTensor[N, 4], "labels": Int64Tensor[N]}` as per the standard in pytorch for detection models. Should this be a choice in the existing dummy dataset?

teddykoker:
Also `VOCDetectionDataModule` won't work right now as they modified an image in the dataset that effects the md5sum of the download. I added a fix to torchvision thats been approved: https://github.com/pytorch/vision/pull/2589

pl_bolts/datamodules/vocdetection_datamodule.py
Borda: well, let's talk about refactoring later...

PR #158, opened 2020-08-15
cpc finetuner

williamFalcon:
adds the finetuner for CPC


pl_bolts/models/self_supervised/ssl_finetuner.py
```diff
@@ -0,0 +1,92 @@
+import pytorch_lightning as pl
```
williamFalcon, line 1: @ananyahjha93  FYI

williamFalcon:
@ananyahjha93 FYI

PR #183, opened 2020-09-05
Fix docstring in `pl_bolts/models/autoencoders/basic_vae/basic_vae_module.py`

akihironitta:
Added an extremely small change to docstring for consistency.


akihironitta:
Looks like this PR isn't necessary anymore, so let me close this PR.

PR #191, opened 2020-09-08
Fix max_depth recursion crash in AsynchronousLoader

DaGuT:
Some people use dict samples with strings, numpy array etc. They are returned from dataloader and are passed to model. Later those are parsed/replaced/removed inside of step/forward function.

Example of such dict (I use it):
`{'image': tensor([WHATEVER]), 'path': [WHATEVER_STRING], 'target': tensor([WHATEVER_NUMPY]), 'meta': {'PatientAge': tensor([WHATEVER]), 'PatientSex': tensor([WHATEVER]), 'StudyDate': tensor([WHATEVER])}, 'mask': tensor([WHATEVER]), 'bboxes': [[tensor([WHATEVER]), tensor([WHATEVER]), tensor([WHATEVER])]]}`

Sadly, async loader was crashing as max depth of recursion was reached (e.g. when it meets string, it just infinitely goes to else in original AsyncLoader). 

**I went to torch.utils.data.default_collate (as it's said that code is based on it), took few lines from there and it works now.** :) 

So it will no longer crash at named_tuples, dict-like objects and strings. Those will all be processed.

# Before submitting

- [ ] Did you make sure to update the docs?   (no need, it's just a small bugfix)

## What does this PR do?
Fixes issue when anything other than torch.tensor containing thing is passed to AsynchronousLoader (more details above)

## PR review    

## Did you have fun?

### DIFF
--- a/pl_bolts/datamodules/async_dataloader.py
+++ b/pl_bolts/datamodules/async_dataloader.py
@@ -1,10 +1,12 @@
 from queue import Queue
 from threading import Thread
 
+import re
+
 import torch
 from torch.utils.data import DataLoader
 from torch.utils.data import Dataset
-
+from torch._six import container_abcs, string_classes, int_classes
 
 class AsynchronousLoader(object):
     """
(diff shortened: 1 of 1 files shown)


DaGuT:
Small note:
Idk if it is related to the patch, but with AsyncLoader it works almost twice slower than without.

PR #247, opened 2020-09-23
U-net implementation 

annikabrundyn:
Adapting this model from an example in the Lightning repo.

Needs to be tested on Kitti dataset which I've submitted a separate PR for.

### DIFF
--- a/pl_bolts/datamodules/__init__.py
+++ b/pl_bolts/datamodules/__init__.py
@@ -1,5 +1,5 @@
 from pl_bolts.datamodules.async_dataloader import AsynchronousLoader
-from pl_bolts.datamodules.dummy_dataset import DummyDetectionDataset
+from pl_bolts.datamodules.dummy_dataset import DummyDataset, DummyDetectionDataset
 
 try:
     from pl_bolts.datamodules.binary_mnist_datamodule import BinaryMNISTDataModule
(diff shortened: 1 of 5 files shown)


oke-aditya:
Very Nice! U-Net was one of the major models. 
U-Net can do semantic segmentation as well as work as convolutional autoencoder. 

So where does this model exactly go into? Into `pl_bolts.models.segmentation` or `pl_bolts.models.autoencoders`
or simply superuseful models go to `pl_bolts.models.vision`

The [original paper](https://arxiv.org/abs/1505.04597) introduced this for medical imaging using semantic segmentation. So it is well known for this.

There are weights available for torch hub as well. [Here](https://github.com/mateuszbuda/brain-segmentation-pytorch) is the repo. This has an implementation of dice loss as well. I'm unsure if it is available in bolts/lightning, and might be good to support.

The implementation is done is very very nice.

Borda (changes requested):
can we pls add simple tests for this U-Net?

annikabrundyn:
> can we pls add simple tests for this U-Net?

Yes! Will add simple tests for U-net today. I've also got the segmentation model that uses U-net as backbone and Kitti datamodule that we can add once this is merged.

Borda:
> I've also got the segmentation model that uses U-net as backbone and Kitti datamodule that we can add once this is merged.

cool, then we could upload the weights to S3 :]

annikabrundyn:
> can we pls add simple tests for this U-Net?

I've added a super simple test just to check that the forward gives the expected shape. Once we merge the Kitti datamodule, I'll add the segmentation model (which uses this Unet architecture) and a test to check it's performance on the dataset

Let me know if you'd like me to add any other tests!

annikabrundyn:
> So where does this model exactly go into? Into `pl_bolts.models.segmentation` or `pl_bolts.models.autoencoders`
> or simply superuseful models go to `pl_bolts.models.vision`

This is a good question - this is basically just the UNet architecture so I'm adding a semantic segmentation model that uses this as a backbone. But yes - it's a useful model for lots of different applications so also wasn't sure where exactly it belongs! And thanks Aditya 😄

PR #248, opened 2020-09-23
Kitti Datamodule

annikabrundyn:
adding Kitti dataset and datamodule for semantic segmentation

note - this requires downloading the dataset first from [this link](http://www.cvlibs.net/datasets/kitti/eval_semseg.php?benchmark=semantics2015)


annikabrundyn:
Hey!
I'm trying to write a test for this datamodule but I'm not sure how to write a general test given that you have to first download the data from the website and then provide the path to the datamodule. 
Any chance you could help share an example of how I should implement this test? :)

Borda (commented):
Can we have a test for this datamodule?

PR #250, opened 2020-09-24
Adds Contibuting.md file

### LINKED ISSUE #212: Create a Contributing.md file
## 🚀 Feature

Just a simple `.md` or `.rst` file which will help people to get going while contributing.

There are a lot of stuff which people keep stumbling, e.g. linting, docs, tests, CI, etc. which are not actually known at the time of PR for newcomers.

Also, we should have clarification about models since this repo is for models mainly

As discussed, what models should go in pl-bolts is also a major question that we can answer in this file.

1. Many papers claim SOTA and to really reproduce results is terribly hard.
2. Less SOTA but more used features are too necessary, e.g. mobilenet backbones in detection models.
3. API should remain consistent, we cannot keep say Pascal VOC format support for FRCNN and then COCO for Detr, it will create inconsistency for the user.
4. Open source licensing, not all are highly permissive like MIT, some are under Apache or GPL.
5. Support for old models, 
E.g. we know that Deep Lab does better than U-Net or LinkNet, do we still add them as they were popular and used before?
E.g. Efficient-Det does better than FRCNN but it might be wise to support both as both have some caveats at few places.
E.g. If we add Retina Net (probably with torchvision very soon), It is a very good single-stage detector, we might not need to use SSD detectors, they aren't so good.

New SOTA models keep popping in every field, (every year rather) how to keep supporting and deprecating.

6. Production support. Pretrained weights are sometimes hard to train and find. Models and implementations are SOTA but weights aren't, what do we do then ?

Feel free to add points further too.

cc @Borda @williamFalcon @ananyahjha93 @nateraw

oke-aditya:
# Before submitting
Closes #212 

## What does this PR do?

Adds `contributing.md` file as discussed. Most contributing file is same as lightning. 
Line 97 introduces an extra section of `New Models` which can be added.

Please let me know what needs to be added more or removed from the current file.

Maybe one small edit in ISSUE_TEMPLATE is needed to point to this contributing guideline.

## PR review    
Need review from all 👍 

## Did you have fun?
Had fun, but no coding here. But next time definitely will code.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -31,7 +31,7 @@ As a researcher, you can't have any part of your code going wrong. So, make thor
 
 #### Interoperability
 
-Have a favorite feature from other libraries like fast.ai or transformers? Those should just work with lightning as well. Grab your favorite model or learning rate scheduler from your favorite library and run it in Lightning.
+Pytorch Lightning bolts is highly interoperable with PyTorch Lightning and PyTorch.
 
 ---
 
@@ -87,9 +87,15 @@ Please raise an issue before adding a new model. Please let us know why the part
 
 ### Test cases:
 
-Want to keep Lightning healthy? Love seeing those green tests? So do we! How to we keep it that way? We write tests! We value tests contribution even more than new features.
+Want to keep Lightning Bolts healthy? Love seeing those green tests? So do we! How to we keep it that way? We write tests! We value tests contribution even more than new features.
 
-Most of the tests in PyTorch Lightning train a trial MNIST model under various trainer conditions (ddp, ddp2+amp, etc...). The tests expect the model to perform to a reasonable degree of testing accuracy to pass. Want to add a new test case and not sure how? [Talk to us!](https://join.slack.com/t/pytorch-lightning/shared_invite/zt-f6bl2l0l-JYMK3tbAgAmGRrlNr00f1A)
+Tests are written using [pytest](https://docs.pytest.org/en/stable/). Tests in PyTorch Lightning bolts train model on a datamodule. Datamodule is lightning abstraction of representing dataloader and dataset. Model is checked by simply calling `.fit()` function over the datamodule.
+
+Along with these we have tests for losses, callbacks and transforms as well.
+
+Have a look at sample tests [here](https://github.com/PyTorchLightning/pytorch-lightning-bolts/tree/master/tests).
+
+Want to add a new test case and not sure how? [Talk to us!](https://join.slack.com/t/pytorch-lightning/shared_invite/zt-f6bl2l0l-JYMK3tbAgAmGRrlNr00f1A)
 
 ---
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,324 @@
...
+3. Submit a PR! We recommend test driven approach to adding new features as well:
+
+   - Write a test for the functionality you want to add.
+   - Write the functional code until the test passes.
+
+4. Add/update the relevant tests!
+
+- [This PR](https://github.com/PyTorchLightning/pytorch-lightning/pull/2671) is a good example for adding a new metric, and [this one for a new logger](https://github.com/PyTorchLightning/pytorch-lightning/pull/2721).
+
+### New Models:
```
oke-aditya, line 97: This is the only new section

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,324 @@
...
+
+1. Models which are standard baselines
+2. Whose results are reproduced properly either by us or by authors.
+3. Top models which are not SOTA but highly cited for production usage / for other uses. (E.g. Moible BERT, MobileNets, FBNets).
+4. Do not reinvent the wheel, natively support torchvision, torchtext, torchaudio models.
+5. Use open source licensed moodels.
+
+Please raise an issue before adding a new model. Please let us know why the particular model is important for bolts. There are tons of models that keep coming. It is very difficult to support every model.
+
+### Test cases:
```
oke-aditya, line 109: I guess this should be edited. Let me know what should go in here if needed.

Borda (changes requested):
it seems that there not many specific sections and some text is very unrelated to Bolts...
I have removed some but could you please prune it more and refer to the PL contributing as needed, it would be for us much easier to maintain only one main contribution guide and in other "supplementary" projects have only specific updates

oke-aditya:
Absolutely it isn't good to repeat and it might become too much read for user. I will try to prune further.

Do let me know how tests are written here which are different from lightning.

Borda:
> Do let me know how tests are written here which are different from lightning.

the tests are the same, just moreover we shall add testing scripts with some arguments parsing...

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,122 @@
...
+If you use some third-party implementation, all such blocks/functions/modules shall be properly referred and if possible also agreed by code's author. For example - `This code is inspired from http://...`.
+In case you adding new dependencies, make sure that they are compatible with the actual PyTorch Lightning license (ie. dependencies should be _at least_ as permissive as the PyTorch Lightning license).
+
+### Question & Answer
+
+1. **How can I help/contribute?**
+
+   All help is extremely welcome - reporting bugs, fixing documentation, adding test cases, solving issues and preparing bug fixes. To solve some issues you can start with label [good first issue](https://github.com/PyTorchLightning/pytorch-lightning/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) or chose something close to your domain with label [help wanted](https://github.com/PyTorchLightning/pytorch-lightning/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22). Before you start to implement anything check that the issue description that it is clear and self-assign the task to you (if it is not possible, just comment that you take it and we assign it to you...).
```
Borda, line 114: update these links to Bolts

oke-aditya: Updated 👍

Borda (approved):

PR #253, opened 2020-09-25
fix negative samples in nt_xent loss

### LINKED ISSUE #252: Little Bug in calculation of negative samples in nt_xent loss
## 🐛 Bug

I think there is a error in the calculation of the [nt_xent loss](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/c60e142905cf44086d5c3e1ed844953730668f98/pl_bolts/losses/self_supervised_learning.py#L8) for self-supervised learning . 

When calculating the indices of negative samples:

`mask = ~torch.eye(n_samples, device=sim.device).bool()`

the indices of positive samples are calculated and then negated. However only the diagonal is considered here, which contains the Self-correlations, thus simply only ones. The true positive samples are located in the right upper and left lower diagonal. That means, simply said, instead of 

loss = -log(pos/neg), 

loss=-log(pos/(pos+neg))

 is calculated. 

In other words, the tensor "neg" should contain only 2N * 2(N-1) samples, because there are exactly 2(N-1) negative samples and one positive sample per data point. In the current code the tensor contains 2N * (2N-1) samples.

I will submit a pull request referring to this issue with an easy solution.

Best, 

Temesgen

Alokaado:
## What does this PR do?
Fixes #252.

### DIFF
--- a/pl_bolts/losses/self_supervised_learning.py
+++ b/pl_bolts/losses/self_supervised_learning.py
@@ -18,8 +18,8 @@ def nt_xent_loss(out_1, out_2, temperature):
 
     # Negative similarity
     diag = torch.eye(n_samples, device=sim.device)
-    right_upper_pos = torch.diagflat(torch.ones(n_samples//2, device=sim.device), offset=n_samples//2)
-    left_lower_pos = torch.diagflat(torch.ones(n_samples//2, device=sim.device), offset=-n_samples//2)
+    right_upper_pos = torch.diagflat(torch.ones(n_samples // 2, device=sim.device), offset=n_samples // 2)
+    left_lower_pos = torch.diagflat(torch.ones(n_samples // 2, device=sim.device), offset=-n_samples // 2)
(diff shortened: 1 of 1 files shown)


pl_bolts/losses/self_supervised_learning.py
```diff
@@ -17,7 +17,10 @@ def nt_xent_loss(out_1, out_2, temperature):
     sim = torch.exp(cov / temperature)
 
     # Negative similarity
-    mask = ~torch.eye(n_samples, device=sim.device).bool()
+    diag = torch.eye(n_samples, device=sim.device)
```
Borda, line 20: can we add a test for this edit...

williamFalcon:
@ananyahjha93

ananyahjha93:
The denominator does contain the positive sample itself, so the calculation checks out to be fine. We do not work with log(pos/neg), but rather log(pos / (pos + sum over neg)). This is not a bug, but rather how the contrastive loss is designed.

PR #256, opened 2020-09-27
[WIP] Adds Detr

### LINKED ISSUE #110: Add DETR
https://arxiv.org/pdf/2005.12872.pdf
https://github.com/facebookresearch/detr

oke-aditya:
# Before submitting

## What does this PR do?
Closes #110 

## PR review    

## Did you have fun?
Yep, This is fun. Finally some time to code 👍

### DIFF
--- a/pl_bolts/models/detection/__init__.py
+++ b/pl_bolts/models/detection/__init__.py
@@ -1,11 +1,3 @@
-__all__ = []
-
-try:
-    from pl_bolts.models.detection.faster_rcnn import FasterRCNN
-    from pl_bolts.models.detection.detr import detr_model
-    from pl_bolts.models.detection.detr import detr_loss
-    from pl_bolts.models.detection.detr import detr_utils
-except ImportError:  # pragma: no-cover
-    pass  # pragma: no-cover
-else:
-    __all__.append('FasterRCNN')
+from pl_bolts.models.detection.faster_rcnn import FasterRCNN
+from pl_bolts.models.detection.detr.detr_model import Detr
+from pl_bolts.models.detection.detr.detr_loss import HungarianMatcher, SetCriterion
--- a/pl_bolts/models/detection/__init__.py
+++ b/pl_bolts/models/detection/__init__.py
@@ -2,6 +2,9 @@ __all__ = []
 
 try:
     from pl_bolts.models.detection.faster_rcnn import FasterRCNN
+    from pl_bolts.models.detection.detr import detr_model
+    from pl_bolts.models.detection.detr import detr_loss
+    from pl_bolts.models.detection.detr import detr_utils
 except ImportError:  # pragma: no-cover
     pass  # pragma: no-cover
 else:
--- a/pl_bolts/models/detection/detr/detr_utils.py
+++ b/pl_bolts/models/detection/detr/detr_utils.py
@@ -1,0 +1,112 @@
+import numpy as np
+import torch.nn as nn
+import torch
+from torchvision.ops.boxes import box_area
+from torch import Tensor
+import torchvision
+
+
+def box_cxcywh_to_xyxy(x):
+    x_c, y_c, w, h = x.unbind(-1)
+    b = [(x_c - 0.5 * w), (y_c - 0.5 * h),
+         (x_c + 0.5 * w), (y_c + 0.5 * h)]
+    return torch.stack(b, dim=-1)
+
+
+def box_xyxy_to_cxcywh(x):
+    x0, y0, x1, y1 = x.unbind(-1)
+    b = [(x0 + x1) / 2, (y0 + y1) / 2,
+         (x1 - x0), (y1 - y0)]
+    return torch.stack(b, dim=-1)
+
+
+# modified from torchvision to also return the union
+def box_iou(boxes1, boxes2):
+    area1 = box_area(boxes1)
+    area2 = box_area(boxes2)
+
+    lt = torch.max(boxes1[:, None, :2], boxes2[:, :2])  # [N,M,2]
+    rb = torch.min(boxes1[:, None, 2:], boxes2[:, 2:])  # [N,M,2]
+
+    wh = (rb - lt).clamp(min=0)  # [N,M,2]
+    inter = wh[:, :, 0] * wh[:, :, 1]  # [N,M]
+
+    union = area1[:, None] + area2 - inter
+
+    iou = inter / union
+    return iou, union
+
+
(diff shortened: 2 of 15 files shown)


oke-aditya:
K there are several code blocks with Detr that can be simplified with October release fo torchvision.
Code in `detr_utils.py` has moved to torchvision master.

Generalized IoU, Bounding boxes conversion, IoU.

This will simplify the code but we need to wait for October.
I have commented out some loss stuff that is needed for segmentation and not for detection. Maybe I can remove that as well.

Let me know thoughts about this. 

This is a slightly different model and DataModule is another concern here which we can think after this.

pl_bolts/models/detection/detr/detr_utils.py
```diff
@@ -0,0 +1,115 @@
+import numpy as np
+import torch.nn as nn
+import torch
+from torchvision.ops.boxes import box_area
```
Borda, line 4: this shall fo to try/catch

pl_bolts/models/detection/__init__.py
```diff
@@ -2,6 +2,9 @@
 
 try:
     from pl_bolts.models.detection.faster_rcnn import FasterRCNN
+    from pl_bolts.models.detection.detr import detr_model
+    from pl_bolts.models.detection.detr import detr_loss
+    from pl_bolts.models.detection.detr import detr_utils
```
Borda, line 7: what is the reason for this wrapping?

Borda (changes requested):

pl_bolts/models/detection/__init__.py
oke-aditya: I didn't get you @Borda. I just tried to Match FRCNN and pass CI tests.
May I know what is the right way of doing this?
Since Detr needed 3 files I moved them to a folder, Let me know the changes.

pl_bolts/models/detection/detr/detr_utils.py
oke-aditya: Moved to try-catch. This file will get cleaned up with next torchvision release.
We would need only `accuracy` and `masks_to_boxes` (if needed).
This codebase has shifted to torchvision and I guess we can use it natively.

pl_bolts/models/detection/__init__.py
Borda: yes, but you wrapped the three imports in try/catch even only one is using `torchvision` so it would be better to isolate this import as much as possible, I 'll check it directly what we can do here...

pl_bolts/models/detection/detr/detr_utils.py
Borda: > This file will get cleaned up with next torchvision release.
> This codebase has shifted to torchvision and I guess we can use it natively in october.

so you say that these metrics will be directly in pyTorch 1.7?

oke-aditya: Yes most should be.
I will patch it in October. It will lessen our code to test and break.

pl_bolts/models/detection/__init__.py
oke-aditya: K. I will isolate it.

williamFalcon: @oke-aditya the try catch is to keep models free of deps :)

oke-aditya: Yes, the only 2 dependencies for `detr` are 
1. `scipy` in `detr_loss` it has no workaround.
2. `torchvision`, for its useful operations, used internally.

oke-aditya: I guess your commit solved this. Really thanks (I will make sure I don't mess it next time)

Borda: Scipy is fine, it is everywhere, the same as NumPy... :]

oke-aditya: Marking as resolved 👍  Thank you for awesome help

oke-aditya:
Almost there. Just for reference This mini-detr or hub detr is from [this](https://github.com/oke-aditya/pytorch_detr) repo. I could train on lightning dummy dataset as well. I just need to add it here.

@justusschock has really nice implementation but with some modification I'm working on that as well.
Plan
1. Get this merged so that we have base reference. This is repo uses orignal FB;s model from torch.hub
2. Get Justus code working with lightning bolts and add that.
3. We can either keep both or choose @justusschock code as it has more features.

oke-aditya:
This is ready for a review !
Detr works on dummy dataset !  I have added tests as well!

This takes original weights from `torch.hub`. I am not quite sure how to add the segmentation model still, but according to the above work plan, the first step seems to be done.

There is lot of code which will go torchvision 0.8. A lot of code in some that need tests but I'm unsure of them.
The code is very dirty and needs refinements please let me know.

justusschock:
@oke-aditya nice work. However, we don't need to implementations of the same thing here. I'd suggest to turn this to a WIP until we merged both codes and then probably open a new PR with the resulting code :)

oke-aditya:
Sure, we can do that. The other idea behind was to get tests and CI actions working for Detr I will get them working in this PR. 
They seem to work fine for this implementation and should probably work fine when we merge both the codes too.

oke-aditya:
Unreleated to this PR but might be something to fixup on other places.
```
FAILED tests/models/test_mnist_templates.py::test_mnist - KeyError: 'loss'
```

oke-aditya:
@Borda I see that base CI is passing with the scipy addition.
But somehow we are still facing the timeout issues. Is this due to matrix build?
I am not sure why it should happen. But there seems some issue.

Borda:
@oke-aditya how is it going here, still wip?

oke-aditya:
Hey @Borda . Acutally this PR might not go to bolts. 
This is complete, It ain't WIP, but @justusschock  has a better implementation which is almost done.
This is very minimal implementation which does model loading from bolts, uses Nested Tensor, etc.

Facing a few bugs [there](https://github.com/justusschock/detr_lightning/pull/3). 
Once those are fixed, we @justusschock can migrate those to bolts.

Borda:
@oke-aditya how is it going here, shall we keep it os close in favor of another work?

oke-aditya:
Let's close this for now. Let's work on organizing codebase and structuring code.
I have a working Detr with Lightning (through torch.hub) maybe we can add it again later..

Borda:
Feel free to re-open, we'll be very happy to have DETR here :]

PR #261, opened 2020-10-04
Explicitly chain exceptions with `raise ... from ...`

akihironitta:
# Before submitting

- [x] Did you make sure to update the docs?   

## What does this PR do?
I'd like to suggest a fix in the way that Python 3's exception chaining is used. I recently submitted a PR and got it merged into PyTorchLightning at https://github.com/PyTorchLightning/pytorch-lightning/pull/3750, and I'd like to apply the fix here as well.

As described in detail in [this article](https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with), exception chaining (PEP 3134) can be used to make exceptions more user-friendly, and in that case, the syntax `raise ... from ...` needs to be used.

When `raise ... from ...` is used, there will be a line saying `The above exception was the direct cause of the following exception:` between tracebacks. However, when implicitly chaining exceptions (meaning when `raise ... from ...` is not used), the message will be `During handling of the above exception, another exception occurred:` which can confuse users.

Specifically, the following should be used in order to explicitly chain exceptions:
```python
try:
    something_which_raises_OldError
except OldError as e:
    raise NewError("A more user-friendly exception message.") from e 
```
instead of:
```python
try:
    something_which_raises_OldError
except OldError:
    raise NewError("A more user-friendly exception message.")
```

## PR review    

## Did you have fun?

### DIFF
--- a/pl_bolts/callbacks/vision/confused_logit.py
+++ b/pl_bolts/callbacks/vision/confused_logit.py
@@ -64,12 +64,12 @@ class ConfusedLogitCallback(Callback):  # pragma: no-cover
         x, y = batch
         try:
             logits = pl_module.last_logits
-        except AttributeError as e:
+        except AttributeError as err:
             m = """please track the last_logits in the training_step like so:
                 def training_step(...):
                     self.last_logits = your_logits
             """
-            raise AttributeError(m)
+            raise AttributeError(m) from err
 
         # only check when it has opinions (ie: the logit > 5)
         if logits.max() > self.min_logit_value:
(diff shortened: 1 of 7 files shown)


Borda:
@akihironitta update/rebased, mind add the same to all ImportExceptions?

akihironitta:
@Borda Thanks for your reponse! It looks good to go! There are `warn()` in `except ModuleNotFoundError` clauses, but those `warn()`ings are not exceptions so they don't need to be chained with `raise ... from ...`, I believe.

Borda (approved):

PR #274, opened 2020-10-13
set min requirements PL 1.0

Borda:
## What does this PR do?

Set min req. PL 1.0

## PR review    

## Did you have fun?


williamFalcon:
why is coverage dropped?

Borda:
> why is coverage dropped?

because we have falling tests, so the complete report is missing and now codecov was only resulting from docstrings inside the package, also ignoring all data modules

Borda:
checked are the failing tests are the same as on the master...

PR #297, opened 2020-10-22
Generelize UNet: custom # of input channels (instead of default 3)

annikabrundyn:


pl_bolts/models/vision/unet.py
```diff
@@ -17,21 +17,23 @@ class UNet(nn.Module):
 
     Args:
         num_classes: Number of output classes required
+        input_channels: Number of channels in input images (default 3)
         num_layers: Number of layers in each side of U-net (default 5)
         features_start: Number of features in first layer (default 64)
         bilinear (bool): Whether to use bilinear interpolation or transposed convolutions (default) for upsampling.
```
akihironitta, line 23:
```suggestion
        bilinear: Whether to use bilinear interpolation or transposed convolutions (default) for upsampling.
```
mind removing ` (bool)` for consistency?

annikabrundyn: good idea - thanks for pointing it out! :)

Issue #251, opened 2020-09-25
Loss Functions for Common Tasks


oke-aditya:
## 🚀 Feature
Frequently re-used losses that can be added to bolts.

### Motivation

Writing Losses is quite repetitive. PyTorch supports losses which are written with deep interoperability with C++ API. But most research losses aren't. Can these go to bolts/lightning (I feel it won't be good in lightning and better in bolts).

These losses are building blocks for other complicates losses as well.

### Pitch

A non-exhaustive and probable list of losses that are not PyTorch but used often.

- [ ] gIoU loss -> Used in Detr (will make porting Detr easier, present in fvcore)
- [ ] focal loss -> Used in RetinaNet (It will come in torchvision as well but we we can have here too for re-use, present in fvcore)
- [ ] smooth l1 loss -> Used in FRCNN and RetinaNet (In torchvision [here](https://github.com/pytorch/vision/blob/master/torchvision/models/detection/_utils.py#L341) and torch but might be helpful)
- [ ] Jsd loss -> Used alternative to CrossEntropy in Classification [here](https://github.com/rwightman/pytorch-image-models/blob/master/timm/loss/jsd.py)
- [ ] Varients of Cross Entropy -> Unsure if they are used often found them [here](https://github.com/rwightman/pytorch-image-models/blob/master/timm/loss/cross_entropy.py)
- [ ] Dice Loss -> Used in U-Net and other segmentation models.
- [ ] Huber loss -> Used in efficnet Det and similar loss. Implemented [here](https://github.com/rwightman/efficientdet-pytorch/blob/2fc4e0e107c969090380d0be1a7ad7552dc00c63/effdet/loss.py#L69)

Unsure of losses in audio and text domains. Someone can add them here as well.

### Alternatives

Wait for them to reach into fvcore or PyTorch. Till then we keep duplicating these code for models.

### Additional context

I might be able to add PRs for these. These look migrations from repositories such as [fvcore](https://github.com/facebookresearch/fvcore), etc. I will cite them as their implementations are used. 
I think that these are likely to be adapted from sources, so unsure of licensing and other policies.

rohitgr7:
I'll try to complete focal loss and smoothing loss this week #121

Cuberick-Orion:
Triplet Loss and its variants can be helpful for metric learning. In case anyone is interested I can help.

PR #347, opened 2020-11-09
Implemented GIoU

briankosw:
## What does this PR do?

Implements Generalized Intersection over Union as mentioned in #251

## Before submitting

## PR review

## Did you have fun?


akihironitta:
@briankosw Thank you for your contribution! Let us know if you need help :]

briankosw:
Hi @akihironitta! I'm trying to write tests for GIoU, but I don't have experience with `unittest` (I'm more used to `pytest`) and it seems that the majority of the tests in this repo weren't fully setup. I'll submit a commit soon, and could you take a look at review it for me?

akihironitta:
@briankosw Sure! Mark this PR as ready for review when you're ready :]

pl_bolts/losses/giou.py
```diff
@@ -0,0 +1,38 @@
...
+    Calculates the generalized intersection over union loss.
+
+    Args:
+        pred: batch of prediction bounding boxes with representation
+              [x_min, y_min, x_max, y_max]
+        target: batch of target bounding boxes with representation
+                [x_min, y_min, x_max, y_max]
+
+    Returns:
+        loss
```
ydcjeff, line 19: rather be `GIoU` loss instead of `loss`?
```suggestion
        GIoU loss
```

tests/losses/test_giou.py
```diff
@@ -0,0 +1,23 @@
...
+
+from unittest import TestCase
+import torch
+
+from pl_bolts.losses.giou import giou_loss
+
+
+class TestGIoULoss(TestCase):
+    def test_complete_overlap(self):
+        pred = torch.tensor([[100, 100, 200, 200]])
```
ydcjeff, line 13: can we test with 10, 20 instead of 100, 200?

ydcjeff (commented):
Great to see GIoU loss in bolts!
Can you add the docs?

tests/losses/test_giou.py
briankosw: So I actually separated the test cases into three separate categories:
- cases where the prediction bbox and target bbox are completely overlapping (i.e. perfect prediction)
- cases where the prediction bbox and target bbox are completely non-overlapping (i.e. completely missed prediction)
- cases where the prediction bbox and target bbox are slightly overlapping

The test case you're suggesting would be slightly overlapping as opposed to completely overlapping. Do you think I should add the new test case you're suggesting or what do you think?

ydcjeff: I see! ok current tests are good

pl_bolts/losses/giou.py
```diff
@@ -0,0 +1,38 @@
...
+                [x_min, y_min, x_max, y_max]
+
+    Returns:
+        GIoU loss
+    """
+    eps = 1e-6
+    x_min = torch.max(pred[:, 0], target[:, 0])
+    y_min = torch.max(pred[:, 1], target[:, 1])
+    x_max = torch.min(pred[:, 2], target[:, 2])
+    y_max = torch.min(pred[:, 3], target[:, 3])
```
akihironitta, line 25: `min` and `max` seem swapped?

akihironitta (changes requested):
Thank you for the swift action! Would you mind having a look?

pl_bolts/losses/giou.py
briankosw: This is actually intended behavior! `x_min` refers to the minimum x value for the intersection between the prediction bbox and the target bbox, and you derive that value by getting the max between the `x_min` of the prediction bbox and the `x_min` of the target bbox.

briankosw: I'd refer you to [Generalized Intersection over Union](https://giou.stanford.edu/), which contains super useful visualizations and explanations for how it works 🙂

akihironitta: I see, then let's keep it as is!

akihironitta (approved):
LGTM!

briankosw:
I'm going to add a little bit more to the docstring and fix that isort failure, but that should be the final touches!

pl_bolts/losses/giou.py
```diff
@@ -0,0 +1,41 @@
...
+    Args:
+        pred: batch of prediction bounding boxes with representation
+              [x_min, y_min, x_max, y_max]
+        target: batch of target bounding boxes with representation
+                [x_min, y_min, x_max, y_max]
+
+    Returns:
+        GIoU loss
+    """
+    eps = 1e-6
```
Borda, line 24: shall this go to constants? this would be depending on if we use AMP (float16) right?

tests/losses/test_giou.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+
+from unittest import TestCase
+
+import torch
+
+from pl_bolts.losses.giou import giou_loss
+
+
+class TestGIoULoss(TestCase):
```
Borda, line 12: can we rather write it as simple `test_complete_overlap` with parameterized inputs and expected results?
it would be easier to debug later as we would si what exact configurations fails

pl_bolts/losses/giou.py
briankosw: Ah good point. Should I extract this out as a global constant or as a passed-in argument?

tests/losses/test_giou.py
briankosw: I'm unfamiliar with `unittest` (I'm used to `pytest`) so I didn't realize you can do that in `unittest` as well!

briankosw: I realized that `pytest` is being used in conjunction with `unittest`, so I just used `pytest` to parameterize the tests.

akihironitta:
Also, please remember to add the summary of your change to the changelog.

tests/losses/test_giou.py
Borda: happy to help if you want, just ping me on slack :]

pl_bolts/losses/giou.py
Borda: well in fact the question is if we really need to use eps as the eps is needed only for zeros decision...

Borda (commented):
pls follow @akihironitta suggestions

annikabrundyn:
This is awesome! Does it make sense to also add this to the PL metrics?

Borda:
> This is awesome! Does it make sense to also add this to the PL metrics?

I would keep it here for a little while and later like any other metric we may move them to PL, we just need to try they are stable and kind of standard :]

ananyahjha93:
@annikabrundyn @briankosw @Borda let's add this to PL as a class metric + the current functional form.

briankosw:
I'd be happy to close this PR and open a new one in PL as soon as I'm given the go ahead 🙂

Borda:
> I'd be happy to close this PR and open a new one in PL as soon as I'm given the go ahead 🙂. If so, I'm assuming #370 and part of #251 should be migrated over as well?

hi, the point is that we shall merge it here, and later (let's say a few weeks) we add it in PL and eventually drop from here

briankosw:
I think this PR is ready for a final review. Would love your reviews @Borda @ydcjeff.

ydcjeff:
Thank you for your work @briankosw 
Is there any difference between this and [this one from torchvision](https://pytorch.org/docs/stable/torchvision/ops.html#torchvision.ops.generalized_box_iou) except that we extract giou from 1 in this?

briankosw:
> Thank you for your work @briankosw
> Is there any difference between this and [this one from torchvision](https://pytorch.org/docs/stable/torchvision/ops.html#torchvision.ops.generalized_box_iou) except that we extract giou from 1 in this?

Ah I didn't realize torchvision has its own implementation. You're correct in that we subtract giou from 1. Should I change it to use torchvision's giou?

ydcjeff:
> > Thank you for your work @briankosw
> > Is there any difference between this and [this one from torchvision](https://pytorch.org/docs/stable/torchvision/ops.html#torchvision.ops.generalized_box_iou) except that we extract giou from 1 in this?
> 
> Ah I didn't realize torchvision has its own implementation. You're correct in that we subtract giou from 1. Should I change it to use torchvision's giou?

I think both are right. in giou loss, we subtract giou from 1. Since torchvision has giou, we only need to 1 - giou to get giou loss... So it can be said this PR is unneeded, Thank you for your work btw.

But let's see core members' thoughts too!

akihironitta:
@briankosw @ydcjeff
Thank you for your work! In my opinion, it's still nice to have the loss here or in PL as the current function and also a subclass of `pl.metrics.Metric` like `pl.metrics.Accuracy` as @ananyahjha93 suggested.

> @annikabrundyn @briankosw @Borda let's add this to PL as a class metric + the current functional form.

I think it's reasonable to have the loss in Bolts and later migrate it to PL like @Borda mentioned.

@Borda @ananyahjha93 So, how shall we move on?

Borda:
yes, as we do not have any loss in PL, we shall also think if it is reasonable to include it there... eventually we would move the losses as a sub-package... so lets it land here and later talk about losses as a complete functional block :]

pl_bolts/losses/object_detection.py
```diff
@@ -0,0 +1,40 @@
...
+    Calculates the generalized intersection over union loss.
+
+    It has been proposed in `Generalized Intersection over Union: A Metric and A
+    Loss for Bounding Box Regression <https://arxiv.org/abs/1902.09630>`_.
+
+    Args:
+        pred: batch of prediction bounding boxes with representation
+              [x_min, y_min, x_max, y_max]
+        target: batch of target bounding boxes with representation
+                [x_min, y_min, x_max, y_max]
```
Borda, line 19:
```suggestion
        pred: batch of prediction bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
        target: batch of target bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
```
minor formating

pl_bolts/losses/object_detection.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Returns:
+        GIoU loss
+    """
```
Borda, line 23: pls add doctest example

pl_bolts/losses/object_detection.py
```diff
@@ -0,0 +1,40 @@
+"""
+Generalized Intersection over Union (GIoU) loss (Rezatofighi et. al)
+"""
+
+import torch
+
+
+def giou_loss(pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
```
Borda, line 8: can we define separately `giou` as functional metric and here use as "supplement"?
so we gain more from this addition :]

Borda (commented):
lets add giou as metric and reuse it in loss

pl_bolts/losses/object_detection.py
briankosw: Forgive my ignorance, but how do I do that?

akihironitta: I think @Borda means something like:
https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/58536c260e7877c8c75dfccfaa58acee4fdc2ae1/pl_bolts/datasets/cifar10_dataset.py#L36-L50

briankosw:
It seems that there is no documentation for metrics in pl_bolts. Should I go ahead and create one? @akihironitta

Borda (approved):

Issue #361, opened 2020-11-13
A bug in precision_at_k function that causes the reported top_k accuracy to be down-scaled by batch size


ahmedhshahin:
## 🐛 A bug in precision_at_k function that causes the reported top_k accuracy to be down-scaled by batch size

The mentioned function divides the accuracy by batch_size in [this line](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/70833f0bac73801a4c970a6ba9324b8469efede8/pl_bolts/metrics/aggregation.py#L32) (to calculate mean accuracy for this batch). However, it returns a list. This list we usually calculate the mean over it again in validation_epoch_end, as in [this line](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/70833f0bac73801a4c970a6ba9324b8469efede8/pl_bolts/models/self_supervised/moco/moco2_module.py#L301), which means we divide by the batch size **two times**. Consequently, the reported accuracy is scaled down by the batch size.

A simple solution to this would be by returning the scalar mean accuracy, not a list.

### To Reproduce

1. Go to pl_bolts.models.self_supervised
2. Run MocoV2
3. Monitor val_acc5 or val_acc1

### Expected behavior
Returning the correct accuracy values. What happens with the current implementation is that it returns accuracy/batch_size.

### Environment

 - PyTorch Version (e.g., 1.0): 1.7.0
 - OS (e.g., Linux): Linux
 - How you installed PyTorch (`conda`, `pip`, source): conda
 - Build command you used (if compiling from source):
 - Python version: 3.6.9
 - CUDA/cuDNN version: 10.00
 - GPU models and configuration: RTX 2080
 - Any other relevant information:

annikabrundyn:
Should this not just be added as a PL metric instead of a Bolts metric?

PR #362, opened 2020-11-13
Fix precision_at_k function

ahmedhshahin:
## What does this PR do?
Fixes issue #361 

## Before submitting
- [ ] Did you make sure to update the documentation with your changes?

## PR review

## Did you have fun?

### DIFF
--- a/pl_bolts/metrics/aggregation.py
+++ b/pl_bolts/metrics/aggregation.py
@@ -29,5 +29,5 @@ def precision_at_k(output, target, top_k=(1,)):
         res = []
         for k in top_k:
             correct_k = correct[:k].contiguous().view(-1).float().sum(0, keepdim=True)
-            res.append(correct_k.mul_(100.0 / batch_size))
+            res.append(correct_k.mul_(100.0).mean())
         return res


pl_bolts/metrics/aggregation.py
```diff
@@ -29,5 +29,5 @@ def precision_at_k(output, target, top_k=(1,)):
         res = []
         for k in top_k:
             correct_k = correct[:k].contiguous().view(-1).float().sum(0, keepdim=True)
-            res.append(correct_k.mul_(100.0 / batch_size))
+            res.append(correct_k.mul_(100.0).mean())
```
Borda, line 32: ok, this seems to solve the bug, but not sure if the problem is somewhere else...
cc: @ananyahjha93 @akihironitta

akihironitta:
@ahmedhshahin Sorry for replying very late. Let me clarify the bug...
I ran the following script:
```
import torch
from pl_bolts.metrics import precision_at_k
output = [
    [0.0, 0.0, 1.0],
    [0.1, 0.2, 0.7],
    [0.2, 0.3, 0.5],
]
target = [2, 1, 2]
top_k = (1,2)
output = torch.tensor(output)
target = torch.tensor(target)
print(precision_at_k(output, target, top_k))
```
, and I got the output:
```txt
[tensor([66.6667]), tensor([100.])]
```
which looks fine as is.
After applying your modification,
I get:
```txt
[tensor(200.), tensor(300.)]
```
which seems not right...

Am I misunderstanding the bug/usage of the function, e.g. `output`/`target` shapes to feed into the fucntion...?

ahmedhshahin:
@akihironitta  No worries! Thanks for your reply and test.
Indeed, the function works well in this scenario. But, the problem is that this function is mostly used at the end of the validation epoch (validation_epoch_end function, [line](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/70833f0bac73801a4c970a6ba9324b8469efede8/pl_bolts/models/self_supervised/moco/moco2_module.py#L301)), where we take the average of the result _again_. So, we are dividing by the batch size _two times_:
-  First time is [here](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/2e903c333c37ea83394c7da2ce826de1b82fb356/pl_bolts/metrics/aggregation.py#L32)
-  Second time is [here](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/70833f0bac73801a4c970a6ba9324b8469efede8/pl_bolts/models/self_supervised/moco/moco2_module.py#L301)

Consequently, the resulting val_acc during training is downscaled by the batch size, as we should divide by it only once.

The problem could be solved by avoiding calculating the mean in the validation_epoch_end function, but this is a standard approach in many other metrics, so changing with this specific function might be confusing. Hence why I suggest changing the precision_at_k function instead.

akihironitta:
@ahmedhshahin I'm sorry I didn't read the description in the linked issue well. You have clearly explained the cause of the unexpected behaviour.
> The mentioned function divides the accuracy by batch_size in this line (to calculate mean accuracy for this batch). However, it returns a list. This list we usually calculate the mean over it again in validation_epoch_end, as in this line, which means we divide by the batch size two times.

So, do you still think we should modify `precision_at_k` instead of [`MocoV2`'s `validation_epoch_end`](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/27855d002405967e8f31d4b33fec523d290e2a3a/pl_bolts/models/self_supervised/moco/moco2_module.py#L299)?

ahmedhshahin:
@akihironitta Thanks for your reply and test.
Actually, I made several tests again and checked your documentation. I think the problem was on my side, I probably made a mistake with the target shape and was using an older version of the function, v0.2.5, the current version seems to be different. I think the current behaviour of the function is correct. Sorry for the inconvenience. I am closing the issue.

PR #371, opened 2020-11-16
Ensure sync across val/test step when using DDP

SeanNaren:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/4693

Lightning behaviour assumes that if no `monitor` key is passed to the checkpoint function, to assume `val_loss` or `checkpoint_on` as the key to save top k models. This means we have to ensure that this is synced correctly on all processes so that they are in sync when tracking the best models.

## PR review

## Did you have fun?

### DIFF
--- a/pl_bolts/models/self_supervised/ssl_finetuner.py
+++ b/pl_bolts/models/self_supervised/ssl_finetuner.py
@@ -106,7 +106,7 @@ class SSLFineTuner(pl.LightningModule):
         acc = self.val_acc(logits, y)
 
         self.log('val_loss', loss, prog_bar=True, sync_dist=True)
-        self.log('val_acc', self.val_acc, sync_dist=True)
+        self.log('val_acc', self.val_acc)
 
         return loss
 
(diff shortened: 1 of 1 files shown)


SeanNaren:
To reduce confusion I can also remove the return loss steps for test/validation, as we're explicitly logging within the function. Let me know if this works @ananyahjha93

Borda (approved):
lgtm

SeanNaren:
@ananyahjha93 and I spoke earlier, need to clean this up a bit.

We have class metrics and normal metrics being used, and are using the val loss to log. This means the sync_dist call doesn't need to be added to the class metric `self.log` call. But then it looks strangely separated:

```python
self.log('val_loss', loss, prog_bar=True, sync_dist=True) # requires sync_dist because not class metric
self.log('val_acc', self.val_acc) # class metric handles sync for us
```

I'm not sure what the resolve here. I've tried to rally for setting sync_dist=True as default, but this has unexpected side effects (what if you want to set val loss reduce to sum but you forgot?)

ananthsub:
Note this came up in #4323 too

SeanNaren:
Hey @ananthsub @Borda what do you think about adding a warning in self.log if this case comes up? Is there a way to make it a one time warning? 

I.e if sync_dist is False for the logged metric and distribution is True, create a warning?

ananyahjha93 (approved):

PR #377, opened 2020-11-18
binary mnist dm fix

annikabrundyn:
Fixes dataset not found error mentioned in issue #353 

Reran the notebook that produced the error and it's now working:

https://colab.research.google.com/drive/1PHwAm48SNojrYbQWPeihmsbWuH-h6KU8#scrollTo=Kh5dXKO4Tr8Y

PR #387, opened 2020-11-20
ci: Fix dataset downloading errors

akihironitta:
## What does this PR do?

As pointed out in https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/377#issuecomment-730193148 by @Borda, the tests try to download datasets, which sometimes fail with the following error:
> UNEXPECTED EXCEPTION: RuntimeError('Failed download from https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz')

### Description of the changes
1. ~It seems that those failing tests are often doctest, so this PR simply removes the doctest from `ci_test-full.yml` as we still have doctest in `ci_test-base.yml`.~
  ~https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/b8ac85154465956b06fd1005b21b071af5493f11/.github/workflows/ci_test-full.yml#L86~
  ~https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/b8ac85154465956b06fd1005b21b071af5493f11/.github/workflows/ci_test-base.yml#L69~
2. ~This PR also includes minor changes in some tests using `LitMNIST` to utilize dataset caching since they currently download and store MNIST datasets in `./` instead of in `./datasets/` (`datadir` fixture).~ See #414.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/.github/workflows/ci_test-full.yml
+++ b/.github/workflows/ci_test-full.yml
@@ -85,6 +85,8 @@ jobs:
         ls -lh
         echo =======
         ls -lh datasets
+        echo =======
+        ls -lh datasets/*
 
     - name: Tests
       run: |
@@ -97,6 +99,8 @@ jobs:
         ls -lh
         echo =======
         ls -lh datasets
+        echo =======
+        ls -lh datasets/*
 
     - name: Upload pytest test results
       uses: actions/upload-artifact@v2
--- a/.github/workflows/ci_test-full.yml
+++ b/.github/workflows/ci_test-full.yml
@@ -102,8 +102,6 @@ jobs:
         echo =======
         ls -lh datasets
         echo =======
-        ls -lh datasets/*
-        echo =======
         find . -name "MNIST"
         find . -name "mnist"
         find . -name "CIFAR"
--- a/.github/workflows/ci_test-full.yml
+++ b/.github/workflows/ci_test-full.yml
@@ -81,6 +81,7 @@ jobs:
         key: pl-datasets-${{ hashFiles('tests/conftest.py') }}
 
     - name: Check cached datasets
+      shell: bash
       run: |
         ls -lh
         echo =======
@@ -95,6 +96,7 @@ jobs:
         coverage xml
 
     - name: Check cached datasets
+      shell: bash
       run: |
         ls -lh
         echo =======
--- a/.github/workflows/ci_test-full.yml
+++ b/.github/workflows/ci_test-full.yml
@@ -86,6 +86,8 @@ jobs:
         ls -lh
         echo =======
         ls -lh datasets
+        echo ======
+        if command -v asdf >/dev/null; then du datasets; fi
         # Commented out because the below lines takes too long on macOS and Windows for unknown reason.
         #echo =======
         #find / -iname "MNIST" 2> /dev/null || echo mnist not found
@@ -103,6 +105,8 @@ jobs:
         ls -lh
         echo =======
         ls -lh datasets
+        echo =======
+        if command -v asdf >/dev/null; then du datasets; fi
         # Commented out because the below lines takes ages on macOS and Windows for unknown reason.
         #echo =======
         #find / -iname "MNIST" 2> /dev/null || echo mnist not found
--- a/.github/workflows/ci_test-full.yml
+++ b/.github/workflows/ci_test-full.yml
@@ -85,7 +85,12 @@ jobs:
       run: |
         ls -lh
         echo =======
-        ls -lh datasets
+        #ls -lh datasets
+        echo =======
+        find . -name "MNIST"
+        find . -name "mnist"
+        find . -name "CIFAR"
+        find . -name "cifar"
 
     - name: Tests
       run: |
@@ -98,7 +103,7 @@ jobs:
       run: |
         ls -lh
         echo =======
-        ls -lh datasets
+        #ls -lh datasets
         echo =======
         find . -name "MNIST"
         find . -name "mnist"
--- a/.github/workflows/ci_test-full.yml
+++ b/.github/workflows/ci_test-full.yml
@@ -88,7 +88,7 @@ jobs:
         #ls -lh datasets
         echo =======
         find / -iname "MNIST" 2> /dev/null
-        find / -iname "CIFAR" 2> /dev/null
+        find / -iname "CIFAR10" 2> /dev/null
 
     - name: Tests
       run: |
@@ -104,7 +104,7 @@ jobs:
         #ls -lh datasets
         echo =======
         find / -iname "MNIST" 2> /dev/null
-        find / -iname "CIFAR" 2> /dev/null
+        find / -iname "CIFAR10" 2> /dev/null
 
     - name: Upload pytest test results
       uses: actions/upload-artifact@v2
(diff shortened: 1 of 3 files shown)


akihironitta:
> It seems that those failing tests are often doctest

This is a false statement... Some tests are still failing other than doctest.

Borda:
> > It seems that those failing tests are often doctest
> 
> This is a false statement... Some tests are still failing other than doctest.

I agree with @akihironitta the point why doctests may seem to be failing more is they are executed before other tests in the tests folder :]

akihironitta:
@Borda I actually have no idea why the tests still fail even after removing the doctests from the full testing CI... e.g. 
> tests/models/test_autoencoders.py::test_vae[cifar10] FAILED

Is there anything else to do here?

.github/workflows/ci_test-full.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     - name: Tests
       run: |
         # tox --sitepackages
-        coverage run --source pl_bolts -m py.test pl_bolts tests -v --junitxml=junit/test-results-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.requires }}.xml
+        coverage run --source pl_bolts -m pytest tests -v --junitxml=junit/test-results-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.requires }}.xml
```
Borda, line 86:
```suggestion
        coverage run --source pl_bolts -m pytest pl_bolts tests -v --junitxml=junit/test-results-${{ runner.os }}-${{ matrix.python-version }}-${{ matrix.requires }}.xml
```
we need to keep testing package e.g. for imports

Borda (commented):
can we just temporary in one run print what is content of the Dataset folder so see if it restored correctly and eventually fro each DataModule show the source folder...

akihironitta:
In docstrings of datasets/datamodules, we're not using `data_dir`, so the doctests try to download and store datasets in `./` instead of using cached datasets in `./datasets/`.

For example, the following test command tries to download and save `./CIFAR10/` instead of reusing cached one in `./datasets/CIFAR10/` which slows down the tests and may be the cause of CI failure.

`coverage run --source pl_bolts -m pytest pl_bolts/datasets/cifar10_dataset.py -v`

akihironitta:
Not sure why the tests on windows tend to fail very often...

akihironitta:
Might be off topic, but I just realised that `CI full testing / pytest (ubuntu-20.04, 3.7, latest)` particularly tend to fail with the error:
```
/home/runner/work/_temp/5ef79e81-ccef-44a4-91a6-610886c324a6.sh: line 2:  1855 Killed                  coverage run --source pl_bolts -m pytest pl_bolts tests --exitfirst -v --junitxml=junit/test-results-Linux-3.7-latest.xml
Error: Process completed with exit code 137.
```

akihironitta:
> Not sure why the tests on windows tend to fail very often...

Seems the last number of ci runs on windows went good without any problems...

.github/workflows/ci_test-full.yml
```diff
@@ -80,12 +80,30 @@ jobs:
...
         coverage xml
 
+    - name: Check cached datasets
+      shell: bash
+      run: |
+        ls -lh
+        echo =======
+        ls -lh datasets
+        echo =======
+        ls -lh datasets/*
```
Borda, line 105:
```suggestion
        echo =======
        ls -lh datasets/*
        echo =======
        find . -name "MNIST"
        find . -name "mnist"
        find . -name "CIFAR"
        find . -name "cifar"
```

.github/workflows/ci_test-full.yml
```diff
@@ -80,12 +80,35 @@ jobs:
...
```
Borda, line 105:
```suggestion
```

.github/workflows/ci_test-full.yml
```diff
@@ -80,12 +80,33 @@ jobs:
...
         key: pl-datasets-${{ hashFiles('tests/conftest.py') }}
 
+    - name: Check cached datasets
+      shell: bash
+      run: |
+        ls -lh
+        echo =======
+        ls -lh datasets
+        echo =======
+        ls -lh datasets/*
```
Borda, line 90:
```suggestion
```

Borda:
it is strange that now all "latest" are failing with CIFAR, but for "minimal" there is no folder CIFAR anywhere in the root...

Borda:
@akihironitta how is it going, can I help somehow? 🐰

akihironitta:
@Borda I was trying to reproduce the error, `RuntimeError('Failed download...)` and see the content of the directory `datasets/` when the error occurs, but the error didn't occur at all for the last tens of runs...

Now, I'll be making several empty commits again to run the tests in parallel to reproduce the error and to see if the cifar10 is correctly cached.

akihironitta:
Now, all say `Cache not found` (even for pip):
- Windows: https://github.com/PyTorchLightning/pytorch-lightning-bolts/runs/1493878834
  - `Cache not found for input keys: pl-datasets-fbdc26bd0289b1874ad7f007a1c3a48f55f391638200617f93e5a8d4ea0e1b09`
- macOS: https://github.com/PyTorchLightning/pytorch-lightning-bolts/runs/1493878712
  - `Cache not found for input keys: pl-datasets-c2e5ea217ff9ce12e555cb270534c632b65e872a766d76b2039b42581f25230a`
- Ubuntu: https://github.com/PyTorchLightning/pytorch-lightning-bolts/runs/1493878672
  - `Cache not found for input keys: pl-datasets-c2e5ea217ff9ce12e555cb270534c632b65e872a766d76b2039b42581f25230a`
  - (`Cache not found for input keys: Linux-pip-py3.7-latest-b7277d5c0e3f381aa4d3321085b48f34312e2047f495a020ed5c6c7e758d899e-, Linux-pip-py3.7-latest-`)

---
This is only a side note, but it is strange that on runs on the same commit 30a7f6d, the hash value on windows is different from the ones on macOS and Ubuntu.

akihironitta:
`datasets/` is only 1.1GB (confirmed locally with `du -hs datasets`), so caches should not exceed the GitHub limit which is 5GB.  
> A repository can have up to 5GB of caches. Once the 5GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.

https://github.com/actions/cache#cache-limits

akihironitta:
It seems ci on the above commit (03fdeab) managed to find the caches. I don't understand why the runs on 30a7f6d and on other commits weren't able to find the caches even though the hash value for the caches was the same because `tests/conftest.py` has never been modified on this PR.

Borda:
looks strange, they say that GH actions are still in development but as much...  :]

akihironitta:
> looks strange, they say that GH actions are still in development but as much... :]

@Borda Right... so what else do you think we can do here? I don't think I can solve this problem because I cannot reproduce the error `UNEXPECTED EXCEPTION: RuntimeError('Failed download...')` for quite a number of runs on GitHub Actions in this branch...

Also, shall we try to fix the below error which happens very often branch in #409? (All the runs on the two commits right above this comment failed due to this.)
```
/home/runner/work/_temp/5ef79e81-ccef-44a4-91a6-610886c324a6.sh: line 2:  1855 Killed                  coverage run --source pl_bolts -m pytest pl_bolts tests --exitfirst -v --junitxml=junit/test-results-Linux-3.7-latest.xml
Error: Process completed with exit code 137.
```

PR #389, opened 2020-11-20
fix last batch index error reinforce

### LINKED ISSUE #381: Bug in reinforce_model causing Index Error: Size mismatch
## 🐛 Bug
When using Reinforce in reinforce_model.py as mentioned, it leads to a size mismatch IndexError  in the last batch of the first epoch. 

#### Code sample
Ran as suggested in [source](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/fa5a9445e9061fd914722b4ef4af461c4135a423/pl_bolts/models/rl/reinforce_model.py#L41):
```
from pl_bolts.models.rl.reinforce_model import Reinforce
from pytorch_lightning import Trainer
model = Reinforce("CartPole-v0")
trainer = Trainer()
trainer.fit(model)
```

#### Error
The error occurs in the last batch of the first epoch.
```
~\AppData\Local\Programs\Miniconda3\envs\torch\lib\site-packages\pl_bolts\models\rl\reinforce_model.py in loss(self, states, actions, scaled_rewards)
    207         # policy loss
    208         log_prob = log_softmax(logits, dim=1)
--> 209         log_prob_actions = scaled_rewards * log_prob[range(self.batch_size), actions]
    210         loss = -log_prob_actions.mean()
    211 

IndexError: shape mismatch: indexing tensors could not be broadcast together with shapes [8], [4]
```

### Environment

 - PyTorch Version (e.g., 1.0): 1.6.0
 - OS (e.g., Linux): Windows
 - How you installed PyTorch (`conda`, `pip`, source): conda
 - Python version: 3.7.8 

###  Additional Info
I can submit a PR to fix this. The error is occurring because in the loss method (line 209), self.batch_size is being referred to and the size of the last batch is less than self.batch_size. Using the length of the sample instead can fix this

sid-sundrani:
## What does this PR do?

This PR replaces getting batch size from self when calculating loss with using the length of the tensor, fixing an IndexError in [Reinforce](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/pl_bolts/models/rl/reinforce_model.py)

The error was an IndexError (size mismatch)  occurring on the last batch of the epoch because tensor size wasn't equal to `self.batch_size`. The number of samples in the dataset may not be divisible by `self.batch_size`, because trajectory collection here is done based on the number of episodes rather than the number of samples. Rather than fixing the error using `drop_last=True` in dataloader, this approach uses all trajectory data 

Fixes #381 

## Before submitting
- [x] Did you make sure to update the documentation with your changes? _No changes_

## Did you have fun?

### DIFF
--- a/pl_bolts/models/rl/reinforce_model.py
+++ b/pl_bolts/models/rl/reinforce_model.py
@@ -215,7 +215,7 @@ class Reinforce(pl.LightningModule):
 
         # policy loss
         log_prob = log_softmax(logits, dim=1)
-        log_prob_actions = scaled_rewards * log_prob[range(self.batch_size), actions]
+        log_prob_actions = scaled_rewards * log_prob[range(len(log_prob)), actions]
         loss = -log_prob_actions.mean()
 
         return loss


akihironitta:
@sid-sundrani Seems you've completed this PR... Make sure to mark this PR ready for review when it's ready :]

akihironitta (approved):
@sid-sundrani LGTM! Thank you for your contribution and swift action :]

Borda (approved):

Borda:
may we add also a test for this issue you have described?

sid-sundrani:
@Borda Hmm okay. Considering this issue, we'd want a test to check if there is an error when current batch size isn't equal to given batch size when calculating the loss? If we set `num_batch_episodes=1`  and `batch_size=1000` (batch_size larger than the number of samples in the first episode), reinforce would only run for one episode and the size of its only training batch won't be equal to the given batch size. Thus it would check if the loss function is compatible with batch size !=self.batch_size. 

If we set those parameters then would running something like this in [TestReinforce](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/tests/models/rl/unit/test_reinforce.py) work?

` def test_loss_epoch_end(self):
    """Test the reinforce loss function"""
    for batch_states, batch_actions, batch_qvals in  self.rl_dataloader:
        loss = self.model.loss(batch_states, batch_actions, batch_qvals)
    `

Borda:
@sid-sundrani mind send a PR with adding test so we can talk about it there...
cc: @tchaton

sid-sundrani:
@Borda Sure. #402

PR #407, opened 2020-11-25
SimSiam

zlapp:
## What does this PR do?

Implement https://arxiv.org/pdf/2011.10566v1.pdf 
Largely based on https://github.com/lucidrains/byol-pytorch extension of BYOL to support SimSiam.
I used pl-bolts BYOL implementation as a reference.
Colab gist for testing on cifar-10 https://gist.github.com/zlapp/c35b8c97d4f6537f21aa07bbc37959c9
Discussed on slack channel https://pytorch-lightning.slack.com/archives/C010PRC9M2R/p1606329394008100

Also adds KNN online evaluation callback

## Before submitting
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure your PR does only one thing, instead of bundling different changes together? Otherwise, we ask you to create a separate PR for every change.
- [ ] Did you make sure to update the documentation with your changes?

## PR review

## Did you have fun?


zlapp:
Initial training results on cifar10 comparing SimSiam (orange) to BYOL (blue).

image
image

I am guessing there is a bug in the initial implementation.

zlapp:
Better online accuracies after fix of detach SimSiam (red) to BYOL (blue):
image
image

Noticed SimSiam uses a factor of -1 times cosine similarity vs BYOL which uses a factor of -2

SimSiam loss:
image

BYOL loss:
image

zlapp:
Changing loss factor for SimSiam to -2 (and not dividing by 2) to be like BYOL improved results.
SimSiam (light blue) to BYOL (blue):
image
image

zlapp:
Reached 88% accuracy on CIFAR10 after 800 epochs:

image

Command:
`python simsiam_module.py --dataset cifar10 --optimizer sgd --batch_size 512 --learning_rate 0.03 --max_epochs 800 --weight_decay 0.0005 --arch resnet18 --hidden_mlp 512`

ananyahjha93 (approved):

pl_bolts/models/self_supervised/simsiam/simsiam_module.py
```diff
@@ -0,0 +1,484 @@
...
+
+        # Image 1 to image 2 loss
+        _, z1, h1 = self.online_network(img_1)
+        _, z2, h2 = self.target_network(img_2)
+        loss_a = -1.0 * self.cosine_similarity(h1, z2)
+
+        # Image 2 to image 1 loss
+        _, z1, h1 = self.online_network(img_2)
+        _, z2, h2 = self.target_network(img_1)
+        loss_b = -1.0 * self.cosine_similarity(h1, z2)
```
haideraltahan, line 189: Based on SimSiam's pseudocode and other [implementations](https://github.com/PatrickHua/SimSiam/blob/main/models/simsiam.py), shouldn't there be just one network without a deep copy target network?

zlapp: Believe it is equivalent. For example if you refer to BYOL implementation in pl bolts there is a deep copy here there isn't https://github.com/lucidrains/byol-pytorch

haideraltahan: Nonetheless, great job on the implementation. It might be worth investigating in the future if there is any difference in performance between the two methods. As with this approach, maybe there is more memory usage?

MikkelAntonsen: With a deep copy instead of using the same network twice you use ~3GB gpu extra memory with resnet18. The two versions are about equally fast on cifar10, but I think that's because the gpu is memorybound on the task.
Given that the final performane ends up the same, I guess removing the copied network would be a good idea as it's less wasteful and scales better.

zlapp: Thanks @MikkelAntonsen.
Could you please share the version you ran with plot of results on CIFAR10?

MikkelAntonsen: https://gist.github.com/MikkelAntonsen/d390d128b28b06a7fc6ed9d2e73b6378
In short, functions init_target_network, on_train_batch_end, shared_step and attribute target_network have been removed. Training_step has been updated accordingly.

bilde

zlapp: Thanks for sharing @MikkelAntonsen.
I just pushed a commit with the changes you suggested in the gist.
Great job making the improvements in efficiency while maintaining accuracy.

MikkelAntonsen: I noticed that the SSLOnlineEvaluator does an additional forward pass per train_batch to get embeddings. It would be possible to return the encoder output from `training_step()` and which would be accessible in the `outputs` argument in SSLOnlineEvaluator, AFAIK. This is unfortunate because it couples the implementation of SSLOnlineEvaluator with the networks that uses it. But if we are to reproduce results on imagenet, is reducing the number of forward passes by 1/3 negligible?

zlapp: Hi @MikkelAntonsen I believe this is related more to SSL capabilities in general of pl_bolts so might be better to open a separate issue since this isn't only effecting SimSiam (tagging @ananyahjha93).

Borda:
@zlapp how is it going here, is it still WIP?

zlapp:
> @zlapp how is it going here, is it still WIP?

Hi @Borda, based on the results here  https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/407#discussion_r549110736 I believe the PR is ready to be merged.

akihironitta (commented):
@zlapp Would you mind having a look at https://github.com/zlapp/pytorch-lightning-bolts/pull/1?

wjn0:
@zlapp as a potential user thanks so much for your (+ the PR reviewers!) work, this looks great. I do have one question: the paper notes in appendix B that one difference between SimSiam and BYOL is the bottleneck structure in the predictor. Namely, the hidden dimension of the predictor MLP should be 1/4 of the output dimension. So, for example, with the default prediction space dimension of 256, I think the hidden dim should be 64. This apparently helps with training stability. Do you agree with my reading of the paper? If so, I'm wondering if this should be the default in the bolt as well?

akihironitta:
> @zlapp Would you mind having a look at [zlapp#1](https://github.com/zlapp/pytorch-lightning-bolts/pull/1)?

Could you also have a look at https://github.com/zlapp/pytorch-lightning-bolts/pull/2?

zlapp (commented):
👍

Borda (approved):

summelon:
The same question as @wjn0.
Any update?

pl_bolts/models/self_supervised/simsiam/simsiam_module.py
MaveriQ: @MikkelAntonsen Maybe the byol implementation can also benefit from using only one network (without deepcopy) and using detach() to control the gradient flow?

MikkelAntonsen: I just read the BYOL paper abstract and it seems like the online network and target network use a different set of weights, so I'm not sure how you could share a network. If you look at figure 2 in the paper, it does seem like they use the stop gradient trick, but only for the target network. Do you see any ways to incorporate simsiam ideas into BYOL without the implemention just ending up as simsiam?

PR #446, opened 2020-12-13
[feature] Contribute Grasp-and-Lift EEG Dataset

thavlik:
## What does this PR do?
This adds the Grasp-and-Lift EEG Dataset from [Multi-channel EEG recordings during 3,936 grasp and lift trials with varying weight and friction](https://www.nature.com/articles/sdata201447).

Improvements I made over the raw data (.csv files) include conversion to binary for fast loading with `torch.load()`. This reduces load speed from several minutes to only a few seconds. I am hosting the converted data, which automatically downloads if it is not locally present. In the event that pre-converted files are lost, the functionality for loading and converting the original .csv files is retained.

This PR includes `grasp_and_lift_eeg_dataset.py` without any other changes. The docs for the other datasets seem to be a WIP (unless I'm looking in the wrong place). I'm happy to make whatever adjustments!

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you make sure your PR does only one thing, instead of bundling different changes together? Otherwise, we ask you to create a separate PR for every change.
- [X] Did you make sure to update the documentation with your changes?

## PR review
 - [X] Is this pull request ready for review? (if not, please submit in draft mode)

## Did you have fun?
Yes!

### DIFF
--- a/pl_bolts/datasets/grasp_and_lift_eeg_dataset.py
+++ b/pl_bolts/datasets/grasp_and_lift_eeg_dataset.py
@@ -10,10 +10,10 @@ import torch.utils.data as data
 
 
 def recursive_listdir(path: str, suffix: str) -> list:
-    return [os.path.join(dp, f)
-            for dp, dn, fn in os.walk(os.path.expanduser(path))
-            for f in fn
-            if f.endswith(suffix)]
+    return [
+        os.path.join(dp, f) for dp, dn, fn in os.walk(os.path.expanduser(path))
+        for f in fn if f.endswith(suffix)
+    ]
 
 
 class GraspAndLiftEEGDataset(data.Dataset):
@@ -101,7 +101,8 @@ class GraspAndLiftEEGDataset(data.Dataset):
                  subjects: Optional[List[int]] = None) -> None:
         super(GraspAndLiftEEGDataset, self).__init__()
         if num_samples is None and last_label_only:
-            raise ValueError('last_label_only cannot be used without setting num_samples')
+            raise ValueError(
+                'last_label_only cannot be used without setting num_samples')
         self.num_samples = num_samples
         self.last_label_only = last_label_only
         data_dir = os.path.join(root, 'train' if train else 'test')
@@ -124,7 +125,8 @@ class GraspAndLiftEEGDataset(data.Dataset):
         else:
             self.load_from_bin(bin_files, subjects)
 
-    def load_from_csv(self, csv_files: List[str], subjects: Optional[List[int]]) -> None:
+    def load_from_csv(self, csv_files: List[str],
+                      subjects: Optional[List[int]]) -> None:
         """ Convert and load the data from the original .csv files
         """
         self.X, self.Y = self.compile_bin(csv_files, subjects)
@@ -134,7 +136,8 @@ class GraspAndLiftEEGDataset(data.Dataset):
             for x in self.X:
                 self.total_examples += x.shape[1] - self.num_samples + 1
 
-    def load_from_bin(self, bin_files: List[str], subjects: Optional[List[int]]) -> None:
+    def load_from_bin(self, bin_files: List[str],
+                      subjects: Optional[List[int]]) -> None:
         """ Load the data from existing .csv.bin files
         """
         examples = {}
@@ -146,8 +149,8 @@ class GraspAndLiftEEGDataset(data.Dataset):
                 if subject not in subjects:
                     continue
             is_data = file.endswith('_data.csv.bin')
-            series = file[:-len('_data.csv.bin')
-                          if is_data else -len('_events.csv.bin')]
+            series = file[:-len('_data.csv.bin'
+                                ) if is_data else -len('_events.csv.bin')]
             samples = torch.load(file)
             item = examples.get(series, [None, None])
             item[0 if is_data else 1] = samples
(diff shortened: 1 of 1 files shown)


pl_bolts/datasets/grasp_and_lift_eeg_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+        train: If true, use the train/ directory and load class labels.
+            If false, load the test/ directory (which lacks labels)
+
+        download: If true, download the data if it is not present locally
+
+        num_samples: The number of samples in the returned examples.
+            If None, the the dataset yields the full length trials.
+
+        last_label_only: If true, return only the last sample's labels.
+            Must be used with num_samples.
```
akihironitta, line 43:
```suggestion
    Args:
        root: Path to directory containing train/ and test/ folders
        train: If true, use the train/ directory and load class labels.
            If false, load the test/ directory (which lacks labels)
        download: If true, download the data if it is not present locally
        num_samples: The number of samples in the returned examples.
            If None, the the dataset yields the full length trials.
        last_label_only: If true, return only the last sample's labels.
            Must be used with num_samples.
```

pl_bolts/datasets/grasp_and_lift_eeg_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+
+    ZIP_URL = 'https://grasplifteeg.nyc3.digitaloceanspaces.com/grasp-and-lift-eeg-detection.zip'
+
+    ZIP_SIZE_BYTES = 980887394
+
+    def __init__(self,
+                 root: str,
+                 train: bool = True,
+                 download: bool = True,
+                 num_samples: int = None,
```
akihironitta, line 92: Please add `from typing import Optional` at the top of this file.
```suggestion
                 num_samples: Optional[int] = None,
```

pl_bolts/datasets/grasp_and_lift_eeg_dataset.py
```diff
@@ -0,0 +1,238 @@
...
+                 last_label_only: bool = False):
```
akihironitta, line 93:
```suggestion
                 last_label_only: bool = False) -> None:
```

akihironitta (commented):
@thavlik Thank you for your contribution! Would you mind adding return types to all functions?
@PyTorchLightning/bolts-contributors Should we have this dataset in Bolts or not?

thavlik:
Github is bugging out on me - please ignore the accidental closure. I have made the suggested changes.

thavlik:
Which signature is preferred for optional return values?

```python
def compile_bin(self, csv_files: List[str]) -> Tuple[Tensor, Optional[Tensor]]:
   ...
```
or
```python
def compile_bin(self, csv_files: List[str]) -> Tuple[Tensor, Union[Tensor, None]]:
   ...
```

Borda:
> Which signature is preferred for optional return values?

I think that Python recommends `Optional[Tensor]`

thavlik:
@akihironitta I have numerous motivations for submitting this dataset in particular. OpenNeuro has other EEG datasets, but there are reasons why this one in particular should be included in bolts. I wouldn't describe visualizing 32-channel EEG as intuitive. But if you are willing to accept the one limitation, Grasp & Lift has a lot of good things going for it.

For starters, it's worth justifying the inclusion of *a* 1D dataset in Bolts. Time series classification is an exceptionally common application. However, example code and pre-trained models for 1D data are generally scarce compared to those available for 2D. While there are clear benefits to focusing on 2D when teaching and learning ML, I think a minimum amount of attention to 1D is warranted. Even if you opt not to include this specific dataset, it is my hope that Bolts can similarly assist the community through other means.

For a random split of the data, the difficulty of the classification task is on easier side of intermediate. >90% validation accuracy **above** baseline (outputting only zero) has been easily achievable with a 1D ResNet-style model of only a few million parameters and <96 hours of training on a single 1080 Ti. It is more ambitious than MNIST, but still feasible with last gen's hardware. There are still questions about generalizability to new subjects (small details like electrode placement matter a lot) but among the same subjects, 90% relative accuracy in 72 GPU-hours is doable.

I suspect this dataset's large number of channels imparts great potential for transfer learning. Data from an entirely different domain with fewer channels can be transformed into a 32-channel representation that performs optimally with a model pretrained on this task.

The size (in Gb) of this dataset palatable. For two seconds worth (1000 samples) there are nearly 18 million examples. Models tend to show high training accuracy before the end of the first epoch, suggesting that the amount and signal/noise ratio of the data is more than sufficient for solving the classification task (or at the very least, for a given set of subjects).

Just like with image classifiers, halving the input resolution (via average pooling) increases training performance. This property is notable because it is not granted that all time series datasets will "work" unless they are at full resolution, which can impose memory limitations and thus be useful to fewer people.

Lastly, and perhaps most importantly, I had fun with it! In principle, a robotic arm could be operated with an EEG cap and the classifier I managed to train on budget hardware. That is not nothing.

Perhaps a 32-channel EEG motor task is a bit too esoteric for Bolts, which currently focuses on the "stock" datasets. Is it the goal of Bolts to accomodate every dataset that can meet its standards? I can see it from both sides.

akihironitta:
@thavlik Thank you very much for your detailed comment. Let's add this dataset to Bolts! I think it would be even nicer maybe to have a model which can be trained on this dataset in the future PR. :]

thavlik:
I've added another argument to the constructor that allows one to choose which of the 12 subjects to load. This will be useful for splitting specific subjects across training/validation, as opposed to randomly selecting examples from all subjects. Doing so should increase the difficulty of the task substantially, as small details such as electrode placement matter a lot.

thavlik:
I've been working with this dataset with separate train/test subjects and it is notably more difficult. How important is difficulty when considering a dataset for pl-bolts? I would say training on a random train/test split is low-moderate difficulty, but splitting train and test subjects is *at least* high-moderate difficulty. I haven't had much time to work on it, so it could even be squarely in the "hard" category.

Issue #434, opened 2020-12-08
Add type annotations to the codebase


akihironitta:
## 🚀 Feature
As static type checker `mypy` was added in #433, we can add type annotations to the codebase.

hassiahk:
Any help needed with this issue? I am happy to help. 😄

akihironitta:
@hassiahk Yes! I just added the steps to contribute to the above comment, so it'd be great if you follow them and submit a PR! :]

hassiahk:
@akihironitta, it would be great if you can also add the below to the initial comment so that we can keep track of which ones are done and this will also ensure no two people are clashing with each other's work. Thanks!!

### Sections
- [ ] `pl_bolts.callbacks.*` <---| @ydcjeff 
- [ ] `pl_bolts.datamodules.*`
- [ ] `pl_bolts.datasets.*`
- [ ] `pl_bolts.losses.*`
- [ ] `pl_bolts.metrics.*`
- [ ] `pl_bolts.models.*`
- [ ] `pl_bolts.models.autoencoders.*`
- [ ] `pl_bolts.models.detection.*`
- [ ] `pl_bolts.models.gans.*`
- [ ] `pl_bolts.models.regression.*`
- [ ] `pl_bolts.models.rl.*`
- [ ] `pl_bolts.models.self_supervised.*`
- [ ] `pl_bolts.models.vision.*`
- [ ] `pl_bolts.optimizers.*`
- [ ] `pl_bolts.transforms.*`
- [ ] `pl_bolts.utils.*` <---| @hassiahk 
- [ ] `tests.*`

akihironitta:
@hassiahk Done :]

hassiahk:
@akihironitta, should I wait for https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/450 to get merged? or can I install `mypy` and proceed with this issue?

akihironitta:
@hassiahk Sure, you can start to work on this issue!

briankosw:
Hey @akihironitta, I'd love to help with this :) Should I just pick on and create a PR?

akihironitta:
@briankosw Yes! Thank you!

hassiahk:
Hey @briankosw. Yes, you can pick one and create a PR for it. 😄 Just leave a comment here saying you picked up this, so you won't clash with others.

briankosw:
I guess I'll work on `pl_bolts.datamodules` then!

[...]

PR #455, opened 2020-12-16
Add Type Annotations to pl_bolts/utils/*

hassiahk:
## What does this PR do?

Part of https://github.com/PyTorchLightning/pytorch-lightning-bolts/issues/434. This PR adds type annotations to `pl_bolts.utils.*`

## Before submitting
- [X] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you make sure your PR does only one thing, instead of bundling different changes together? Otherwise, we ask you to create a separate PR for every change.
- [X] Did you make sure to update the documentation with your changes?

## PR review
 - [X] Is this pull request ready for review? (if not, please submit in draft mode)

## Did you have fun?

### DIFF
--- a/pl_bolts/utils/arguments.py
+++ b/pl_bolts/utils/arguments.py
@@ -1,7 +1,7 @@
 import inspect
 from argparse import ArgumentParser, Namespace
 from dataclasses import dataclass
-from typing import Any, Optional
+from typing import Any, Dict, List, Optional
 
 import pytorch_lightning as pl
 
@@ -32,7 +32,7 @@ class LightningArgumentParser(ArgumentParser):
         # args.data -> data args
         # args.model -> model args
     """
-    def __init__(self, *args, ignore_required_init_args=True, **kwargs):
+    def __init__(self, *args: Any, ignore_required_init_args: bool = True, **kwargs: Any):
         """
         Args:
             ignore_required_init_args (bool, optional): Whether to include positional args when adding
@@ -41,10 +41,10 @@ class LightningArgumentParser(ArgumentParser):
         super().__init__(*args, **kwargs)
         self.ignore_required_init_args = ignore_required_init_args
 
-        self._default_obj_args = dict()
-        self._added_arg_names = []
+        self._default_obj_args: Dict[str, List[LitArg]] = dict()
+        self._added_arg_names: List[str] = []
 
-    def add_object_args(self, name, obj):
+    def add_object_args(self, name: str, obj: Any) -> None:
         default_args = gather_lit_args(obj)
         self._default_obj_args[name] = default_args
         for arg in default_args:
@@ -58,7 +58,7 @@ class LightningArgumentParser(ArgumentParser):
                 kwargs["default"] = arg.default
             self.add_argument(f"--{arg.name}", **kwargs)
 
-    def parse_lit_args(self, *args, **kwargs):
+    def parse_lit_args(self, *args: Any, **kwargs: Any) -> Namespace:
         parsed_args_dict = vars(self.parse_args(*args, **kwargs))
         lit_args = Namespace()
         for name, default_args in self._default_obj_args.items():
@@ -72,7 +72,7 @@ class LightningArgumentParser(ArgumentParser):
         return lit_args
 
 
-def gather_lit_args(cls, root_cls=None):
+def gather_lit_args(cls: Any, root_cls: Optional[Any] = None) -> List[LitArg]:
 
     if root_cls is None:
         if issubclass(cls, pl.LightningModule):
@@ -83,7 +83,7 @@ def gather_lit_args(cls, root_cls=None):
             root_cls = cls
 
     blacklisted_args = ["self", "args", "kwargs"]
-    arguments = []
+    arguments: List[LitArg] = []
     argument_names = []
     for obj in inspect.getmro(cls):
 
--- a/pl_bolts/utils/arguments.py
+++ b/pl_bolts/utils/arguments.py
@@ -92,7 +92,7 @@ def gather_lit_args(cls: Any, root_cls: Optional[Any] = None) -> List[LitArg]:
 
         if issubclass(obj, root_cls):
 
-            default_params = inspect.signature(obj.__init__).parameters
+            default_params = inspect.signature(obj.__init__).parameters  # type: ignore
 
             for arg in default_params:
                 arg_type = default_params[arg].annotation
@@ -104,7 +104,7 @@ def gather_lit_args(cls: Any, root_cls: Optional[Any] = None) -> List[LitArg]:
                     arg_types = (arg_type,)
 
                 # If type is empty, that means it hasn't been given type hint. We skip these.
-                arg_is_missing_type_hint = arg_types == (inspect._empty,)
+                arg_is_missing_type_hint = arg_types == (inspect.Parameter.empty,)
                 # Some args should be ignored by default (self, kwargs, args)
                 arg_is_in_blacklist = arg in blacklisted_args and arg_is_missing_type_hint
                 # We only keep the first arg we see of a given name, as it overrides the parents
@@ -113,9 +113,9 @@ def gather_lit_args(cls: Any, root_cls: Optional[Any] = None) -> List[LitArg]:
                 do_skip_this_arg = arg_is_in_blacklist or arg_is_missing_type_hint or arg_is_duplicate
 
                 # Positional args have no default, but do have a known type or types.
-                arg_is_positional = arg_default == inspect._empty and not arg_is_missing_type_hint
+                arg_is_positional = arg_default == inspect.Parameter.empty and not arg_is_missing_type_hint
                 # Kwargs have both a default + known type or types
-                arg_is_kwarg = arg_default != inspect._empty and not arg_is_missing_type_hint
+                arg_is_kwarg = arg_default != inspect.Parameter.empty and not arg_is_missing_type_hint
 
                 if do_skip_this_arg:
                     continue
@@ -124,7 +124,7 @@ def gather_lit_args(cls: Any, root_cls: Optional[Any] = None) -> List[LitArg]:
                     lit_arg = LitArg(
                         name=arg,
                         types=arg_types,
-                        default=arg_default if arg_default != inspect._empty else None,
+                        default=arg_default if arg_default != inspect.Parameter.empty else None,
                         required=arg_is_positional,
                         context=obj.__name__,
                     )
--- a/pl_bolts/utils/arguments.py
+++ b/pl_bolts/utils/arguments.py
@@ -32,11 +32,11 @@ class LightningArgumentParser(ArgumentParser):
         # args.data -> data args
         # args.model -> model args
     """
-    def __init__(self, *args: Any, ignore_required_init_args: bool = True, **kwargs: Any):
+    def __init__(self, *args: Any, ignore_required_init_args: bool = True, **kwargs: Any) -> None:
         """
         Args:
-            ignore_required_init_args (bool, optional): Whether to include positional args when adding
-            object args. Defaults to True.
+            ignore_required_init_args: Whether to include positional args when adding
+            object args. Defaults to ``True``.
         """
         super().__init__(*args, **kwargs)
         self.ignore_required_init_args = ignore_required_init_args
--- a/pl_bolts/utils/semi_supervised.py
+++ b/pl_bolts/utils/semi_supervised.py
@@ -57,8 +57,8 @@ def balance_classes(
     nb_batches = math.ceil(len(Y) / batch_size)
 
     # sort by classes
-    final_batches_x: List[Any] = [[] for i in range(nb_batches)]
-    final_batches_y: List[Any] = [[] for i in range(nb_batches)]
+    final_batches_x: List[list] = [[] for i in range(nb_batches)]
+    final_batches_y: List[list] = [[] for i in range(nb_batches)]
 
     # Y needs to be np arr
     Y = np.asarray(Y)
--- a/pl_bolts/utils/semi_supervised.py
+++ b/pl_bolts/utils/semi_supervised.py
@@ -1,5 +1,5 @@
 import math
-from typing import Any, List, Tuple
+from typing import Any, List, Sequence, Tuple
 
 import numpy as np
 import torch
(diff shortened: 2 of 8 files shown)


hassiahk:
@akihironitta, need help with the below two errors. I am not sure how to fix them.
- `Cannot access "__init__" directly`
- `Module has no attribute "_empty"`

pl_bolts/utils/semi_supervised.py
```diff
@@ -51,19 +53,19 @@ def balance_classes(X: np.ndarray, Y: list, batch_size: int):
...
-    final_batches_y = [[] for i in range(nb_batches)]
+    final_batches_x: List[Any] = [[] for i in range(nb_batches)]
+    final_batches_y: List[Any] = [[] for i in range(nb_batches)]
 
     # Y needs to be np arr
     Y = np.asarray(Y)
 
     # pick chunk size for each class using the largest split
-    chunk_size = []
+    chunk_size_class = []
```
hassiahk, line 63: @akihironitta, `mypy` was giving error if I don't change `chunk_size` to something else because it is `List[int]` and when we do `chunk_size = max(chunk_size)` it is becoming `int`. Let me know if this okay to do.

pl_bolts/utils/semi_supervised.py
```diff
@@ -51,19 +53,19 @@ def balance_classes(X: np.ndarray, Y: list, batch_size: int):
     nb_batches = math.ceil(len(Y) / batch_size)
 
     # sort by classes
-    final_batches_x = [[] for i in range(nb_batches)]
-    final_batches_y = [[] for i in range(nb_batches)]
+    final_batches_x: List[Any] = [[] for i in range(nb_batches)]
+    final_batches_y: List[Any] = [[] for i in range(nb_batches)]
```
hassiahk, line 57: `mypy` was asking me to annotate these two.

akihironitta, line 57: In my opinion, we shouldn't use `Any` where possible, so can we rewrite it as follows?
```suggestion
    final_batches_x: List[List[int]] = [[] for i in range(nb_batches)]
    final_batches_y: List[List[int]] = [[] for i in range(nb_batches)]
```

pl_bolts/utils/semi_supervised.py
akihironitta: I think you're right. How about renaming the variable `chunk_size_class` to `chunk_sizes`?

pl_bolts/utils/arguments.py
```diff
@@ -32,7 +32,7 @@ class LightningArgumentParser(ArgumentParser):
         # args.data -> data args
         # args.model -> model args
     """
-    def __init__(self, *args, ignore_required_init_args=True, **kwargs):
+    def __init__(self, *args: Any, ignore_required_init_args: bool = True, **kwargs: Any):
```
akihironitta, line 35:
```suggestion
    def __init__(self, *args: Any, ignore_required_init_args: bool = True, **kwargs: Any) -> None:
```

pl_bolts/utils/arguments.py
```diff
@@ -41,10 +41,10 @@ def __init__(self, *args, ignore_required_init_args=True, **kwargs):
         super().__init__(*args, **kwargs)
         self.ignore_required_init_args = ignore_required_init_args
 
-        self._default_obj_args = dict()
-        self._added_arg_names = []
+        self._default_obj_args: Dict[str, List[LitArg]] = dict()
```
akihironitta, line 44: This is completely off topic and not a strong opinion, but we should use `{}` instead of `dict()` to instantiate an empty dictionary as I remember `{}` is faster than `dict()`.

@hassiahk Would you be interested in submitting a seperate PR replacing `dict()` with `{}` in the codebase?

pl_bolts/utils/arguments.py
```diff
@@ -32,7 +32,7 @@ class LightningArgumentParser(ArgumentParser):
         # args.data -> data args
         # args.model -> model args
     """
-    def __init__(self, *args, ignore_required_init_args=True, **kwargs):
+    def __init__(self, *args: Any, ignore_required_init_args: bool = True, **kwargs: Any):
         """
         Args:
             ignore_required_init_args (bool, optional): Whether to include positional args when adding
```
akihironitta, line 38: As we have function annotations here, the types of args should be automatically reflected in the docs, so can you remove these from docstrings?
```suggestion
            ignore_required_init_args: Whether to include positional args when adding
```

pl_bolts/utils/semi_supervised.py
```diff
@@ -24,14 +26,14 @@ class Identity(torch.nn.Module):
...
+    def __init__(self) -> None:
         super(Identity, self).__init__()
 
-    def forward(self, x):
+    def forward(self, x: Tensor) -> Tensor:
         return x
 
 
-def balance_classes(X: np.ndarray, Y: list, batch_size: int):
+def balance_classes(X: np.ndarray, Y: list, batch_size: int) -> Tuple[np.ndarray, np.ndarray]:
```
akihironitta, line 36: May we be more specific here?
```suggestion
def balance_classes(X: np.ndarray, Y: List[int], batch_size: int) -> Tuple[np.ndarray, np.ndarray]:
```

pl_bolts/utils/warnings.py
```diff
@@ -1,10 +1,16 @@
...
+from typing import Any, Callable
 
 MISSING_PACKAGE_WARNINGS = {}
 
 
-def warn_missing_pkg(pkg_name: str, pypi_name: str = None, extra_text: str = None, stdout_func=warnings.warn):
+def warn_missing_pkg(
+    pkg_name: str,
+    pypi_name: str = None,
+    extra_text: str = None,
```
akihironitta, line 11:
```suggestion
    pypi_name: Optional[str] = None,
    extra_text: Optional[str] = None,
```

akihironitta (changes requested):
@hassiahk Thank you for your contribution! I left some comments on the changes. Could you have a look? As for your comment below, I'll look into it later...
> @akihironitta, need help with the below two errors. I am not sure how to fix them.
> - `Cannot access "__init__" directly`
> - `Module has no attribute "_empty"`

pl_bolts/utils/arguments.py
hassiahk: Sure, I can do this.

pl_bolts/utils/arguments.py
hassiahk: Sure, can do this as well.

pl_bolts/utils/semi_supervised.py
hassiahk: Yes, I was not sure regarding the type, so keep it as `Any`, will change that now.

pl_bolts/utils/semi_supervised.py
hassiahk: My bad, I think I missed this because it was already annotated.😄

hassiahk: Should I change it to `Sequence[int]` since it can be a `list` or a `tuple`?

pl_bolts/utils/semi_supervised.py
akihironitta: Sorry, what I've just mentioned is completely false statement.  Still, it would be nice to have a more specific type, but we can keep it as is since it seems not obvious...

hassiahk: Okay, sure.

hassiahk:
@akihironitta, I took care of your requested changes but I also had to change `Y -> labels` in `balance_classes` while annotating it as `Sequence[int]` or `List[int]` because `mypy` is throwing below errors:
- `Argument 1 to "len" has incompatible type "int"; expected "Sized"` -> `mypy` is basically saying that you did not take care of `Y` being `None`.
- `Value of type "int" is not indexable` -> Not sure of this.

Another workaround would be to annotate `Y` as plain `Sequence` or `List`. Let me know your thoughts.

pl_bolts/utils/arguments.py
hassiahk: @akihironitta, the entire codebase or just the `pl_bolts.utils.*`?

akihironitta: @hassiahk  Please do to the whole codebase in a single PR :]

pl_bolts/utils/semi_supervised.py
akihironitta: @hassiahk On second thought, I don't think the annotated types are not correct. In the tests, we use the function as the following:
```python
tests/utils/test_semi_supervised.py:9:def test_balance_classes():
tests/utils/test_semi_supervised.py-10-    X = torch.rand(100, 3, 32, 32)
tests/utils/test_semi_supervised.py-11-    c1 = torch.zeros(20, 1)
tests/utils/test_semi_supervised.py-12-    c2 = torch.zeros(20, 1) + 1
tests/utils/test_semi_supervised.py-13-    c3 = torch.zeros(60, 1) + 2
tests/utils/test_semi_supervised.py-14-
tests/utils/test_semi_supervised.py-15-    y = torch.cat([c1, c2, c3], dim=0).int().cpu().numpy().flatten().tolist()
tests/utils/test_semi_supervised.py:16:    (balanced_X, balanced_Y) = balance_classes(X, y, batch_size=10)
```
which suggests we should annotate them differently from the current ones.

akihironitta (changes requested):
@hassiahk Thank you very much for your work!

For `Module has no attribute "_empty"`, let's replace `inspect._empty` with `inspect.Parameter.empty` because both should be equivalent which is defined in [`inspect`](https://github.com/python/cpython/blob/master/Lib/inspect.py). The same issue was also discussed in https://github.com/allenai/allennlp/pull/1191#discussion_r201096345.

For `Cannot access "__init__" directly`, I couldn't find any existing solution, so shall we just skip the line with `# type: ignore`...?

pl_bolts/utils/semi_supervised.py
```diff
@@ -24,46 +26,46 @@ class Identity(torch.nn.Module):
...
+    def __init__(self) -> None:
         super(Identity, self).__init__()
 
-    def forward(self, x):
+    def forward(self, x: Tensor) -> Tensor:
         return x
 
 
-def balance_classes(X: np.ndarray, Y: list, batch_size: int):
+def balance_classes(X: np.ndarray, labels: Sequence[int], batch_size: int) -> Tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
hassiahk: Ah, I did not think it can be `torch.Tensor` or `numpy.ndarray`, I think this will solve the problem.
> (or you think of other types, too?).
It can also be `Tuple`, hence I kept it as `Sequence`

pl_bolts/utils/semi_supervised.py
```diff
@@ -24,46 +26,46 @@ class Identity(torch.nn.Module):
...
     """
     Makes sure each batch has an equal amount of data from each class.
     Perfect balance
 
     Args:
         X: input features
-        Y: mixed labels (ints)
+        labels: mixed labels (ints)
```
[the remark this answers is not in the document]
hassiahk: Okay, sure. I think annotating it as `Union` will solve the problem. If not, I can rename the variables that are inside the function.

pl_bolts/utils/semi_supervised.py
hassiahk: Annotated `X` as `Union[Tensor, np.ndarray]` and `Y` as `Union[Tensor, np.ndarray, Sequence[int]]`.

pl_bolts/utils/semi_supervised.py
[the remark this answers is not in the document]
hassiahk: Annotating it as `Union` solved the problem.

hassiahk:
@akihironitta, made requested changes. Kindly review it once again. Annotated `X` as `Union[Tensor, np.ndarray]` and `Y` as `Union[Tensor, np.ndarray, Sequence[int]]`.

pl_bolts/utils/semi_supervised.py
```diff
@@ -24,14 +26,18 @@ class Identity(torch.nn.Module):
...
 
-    def forward(self, x):
+    def forward(self, x: Tensor) -> Tensor:
         return x
 
 
-def balance_classes(X: np.ndarray, Y: list, batch_size: int):
+def balance_classes(
+    X: Union[Tensor, np.ndarray],
+    Y: Union[Tensor, np.ndarray, Sequence[int]],
```
akihironitta, line 38: I'm still concerned about if the annotated types include all the possibility, but I think these are fine for now.

pl_bolts/utils/semi_supervised.py
```diff
@@ -51,19 +57,19 @@ def balance_classes(X: np.ndarray, Y: list, batch_size: int):
     nb_batches = math.ceil(len(Y) / batch_size)
 
     # sort by classes
-    final_batches_x = [[] for i in range(nb_batches)]
-    final_batches_y = [[] for i in range(nb_batches)]
+    final_batches_x: List[Any] = [[] for i in range(nb_batches)]
+    final_batches_y: List[Any] = [[] for i in range(nb_batches)]
```
akihironitta, line 61: Since these variables are the output of the function, I think they should match with the return type of the function `balance_classes()`:
```python
) -> Tuple[np.ndarray, np.ndarray]:
```
but in order to do that, you might need to change/define other variables, too. We can tighten these types later in another PR, but it'd be nice to do it in this PR. Do you think you can improve typing here?

akihironitta (approved):
@hassiahk Thank you for your work! LGTM :]

pl_bolts/utils/arguments.py
akihironitta: @hassiahk Look forward to reviewing your PR on this :]

pl_bolts/utils/semi_supervised.py
hassiahk: Sure, I will see what I can do. :)

Borda, line 61:
```suggestion
    final_batches_x: List[list] = [[] for i in range(nb_batches)]
    final_batches_y: List[list] = [[] for i in range(nb_batches)]
```

Borda (approved):

pl_bolts/utils/semi_supervised.py
hassiahk: I will use this `balance_classes` and see if we can use other types as well.

PR #469, opened 2020-12-21
Intersection over Union Metric/Loss

### LINKED ISSUE #370: Add Intersection over Union loss
## 🚀 Feature

Since #251 includes GIoU loss, I think having IoU loss only makes sense.

### Pitch

I can implement this pretty quickly since I've been working on GIoU #347.

briankosw:
## What does this PR do?

Implements IoU metric and IoU loss

Fixes #370. Related to #347 and #251.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pl_bolts/metrics/object_detection.py
+++ b/pl_bolts/metrics/object_detection.py
@@ -6,8 +6,8 @@ def iou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
     Calculates the intersection over union.
 
     Args:
-        preds: batch of prediction bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
-        target: batch of target bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
+        preds: an Nx4 batch of prediction bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
+        target: an Mx4 batch of target bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
 
     Example:
 
--- a/pl_bolts/metrics/object_detection.py
+++ b/pl_bolts/metrics/object_detection.py
@@ -1,6 +1,38 @@
 import torch
 
 
+def iou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+    """
+    Calculates the intersection over union.
+
+    Args:
+        preds: batch of prediction bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
+        target: batch of target bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
+
+    Example:
+
+        >>> import torch
+        >>> from pl_bolts.metrics.object_detection import iou
+        >>> preds = torch.tensor([[100, 100, 200, 200]])
+        >>> target = torch.tensor([[150, 150, 250, 250]])
+        >>> iou(preds, target)
+        tensor([[0.1429]])
+
+    Returns:
+        IoU value
+    """
+    x_min = torch.max(preds[:, None, 0], target[:, 0])
+    y_min = torch.max(preds[:, None, 1], target[:, 1])
+    x_max = torch.min(preds[:, None, 2], target[:, 2])
+    y_max = torch.min(preds[:, None, 3], target[:, 3])
+    intersection = (x_max - x_min).clamp(min=0) * (y_max - y_min).clamp(min=0)
+    pred_area = (preds[:, 2] - preds[:, 0]) * (preds[:, 3] - preds[:, 1])
+    target_area = (target[:, 2] - target[:, 0]) * (target[:, 3] - target[:, 1])
+    union = pred_area[:, None] + target_area - intersection
+    iou = torch.true_divide(intersection, union)
+    return iou
+
+
 def giou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
     """
     Calculates the generalized intersection over union.
--- a/pl_bolts/metrics/object_detection.py
+++ b/pl_bolts/metrics/object_detection.py
@@ -19,7 +19,8 @@ def iou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
         tensor([[0.1429]])
 
     Returns:
-        IoU value
+        IoU tensor: an NxM tensor containing the pairwise IoU values for every element in preds and target,
+                    where N is the number of prediction bounding boxes and M is the number of target bounding boxes
     """
     x_min = torch.max(preds[:, None, 0], target[:, 0])
     y_min = torch.max(preds[:, None, 1], target[:, 1])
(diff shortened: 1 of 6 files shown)


oke-aditya:
But IoU and gIoU are available in torchvision. Thoughts ?

pl_bolts/metrics/object_detection.py
```diff
@@ -1,6 +1,38 @@
...
+    y_max = torch.min(preds[:, None, 3], target[:, 3])
+    intersection = (x_max - x_min).clamp(min=0) * (y_max - y_min).clamp(min=0)
+    pred_area = (preds[:, 2] - preds[:, 0]) * (preds[:, 3] - preds[:, 1])
+    target_area = (target[:, 2] - target[:, 0]) * (target[:, 3] - target[:, 1])
+    union = pred_area[:, None] + target_area - intersection
+    iou = torch.true_divide(intersection, union)
+    return iou
+
+
 def giou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
```
oke-aditya, line 36: Even giou is implemented in torchvision

briankosw:
> But IoU and gIoU are available in torchvision. Thoughts ?
> I agree that gIoU loss IoU loss is not available in torchvision (hopefully soon) but for ops, there is tested support.
> 
> cc @akihironitta maybe we are reimplementing these in bolts to avoid dependency ?

The conversation regarding `pl_bolts` having its own implementation of `giou` and `giou_loss` was held in #347. The conclusion was that it's helpful to have a separate version, hence this PR.

oke-aditya:
Hmm, not an issue then but we would need to test to match torchvision's implementation exactly.
E.g. IoU and gIoU in torchvision return a `Tensor[N, 4]` while here we do a little bit different.
Which might be slightly confusing for users.

It can be nice to have drop-in replacement for consistency.  Thoughts ?

cc @akihironitta @Borda

briankosw:
> Hmm, not an issue then but we would need to test to match torchvision's implementation exactly.
> E.g. IoU and gIoU in torchvision return a `Tensor[N, 4]` while here we do a little bit different.
> Which might be slightly confusing for users.

So IoU and GIoU in `torchvision` actually doesn't return an Nx4 tensor, it returns an NxM tensor, where N is the dimension of the first set of boxes and M is the dimension of the second set of boxes. In fact my implementation does the same, and you bring up an excellent point that perhaps the docstring should correctly reflect this fact to the users. In addition, I can further demonstrate this by creating tests that reflect this fact. Does that sound ok?

oke-aditya:
Oh yes, it returns Tensor[N, M] Yes, we can even test on same values as we did in torchvision.
Tests in torchvision are [here](https://github.com/pytorch/vision/blob/master/test/test_ops.py#L838)

pl_bolts/metrics/object_detection.py
```diff
@@ -1,6 +1,38 @@
...
+
+    Returns:
+        IoU value
+    """
+    x_min = torch.max(preds[:, None, 0], target[:, 0])
+    y_min = torch.max(preds[:, None, 1], target[:, 1])
+    x_max = torch.min(preds[:, None, 2], target[:, 2])
+    y_max = torch.min(preds[:, None, 3], target[:, 3])
+    intersection = (x_max - x_min).clamp(min=0) * (y_max - y_min).clamp(min=0)
+    pred_area = (preds[:, 2] - preds[:, 0]) * (preds[:, 3] - preds[:, 1])
```
oke-aditya, line 29: thoughts to have a helper `box_area` function as in torchvision ?

briankosw:
Great suggestion @oke-aditya! How should I go about changing the description for gIoU? Do I create a separate PR?

pl_bolts/metrics/object_detection.py
briankosw: I didn't think a separate `box_area` method was warranted since its use case is quite narrow. Should I go ahead and implement it?

oke-aditya:
I'm Not sure about that. cc @akihironitta :smile:

pl_bolts/metrics/object_detection.py
oke-aditya: Maybe lets leave it for now. Create when we really need it?

akihironitta:
> But IoU and gIoU are available in torchvision. Thoughts?
> I agree that gIoU loss IoU loss is not available in torchvision (hopefully soon) but for ops, there is tested support.
> 
> cc @akihironitta maybe we are reimplementing these in bolts to avoid dependency?

Since Bolts already (heavily?) depends on `torchvision`, I think we can import `torchvision`'s (g)iou and use them here. It's also nice that they should be tested well by the torchvision team. In #347, @briankosw asked if we should have own implementation, but it seems (and I'm sorry) that we missed the question. So, for now, let's hear what others think of this. cc: @PyTorchLightning/bolts-contributors @Borda 

> > Is there any difference between this and [this one from torchvision](https://pytorch.org/docs/stable/torchvision/ops.html#torchvision.ops.generalized_box_iou) except that we extract giou from 1 in this?
> 
> Ah I didn't realize torchvision has its own implementation. You're correct in that we subtract giou from 1. Should I change it to use torchvision's giou?

---

> It would nice to demonstrate with docstring as well. Maybe just say
 
@briankosw Let's improve the description of the loss in the docs. Could you work on this in another PR?

briankosw:
I'll pause work on this PR, and I'll create another PR that improves the description of GIoU loss then.

briankosw:
I'll pause work on this PR then. I already updated the docs for IoU metric as @oke-aditya suggested, so what description should I improve? @akihironitta

pl_bolts/metrics/object_detection.py
```diff
@@ -1,6 +1,39 @@
 import torch
 
 
+def iou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+    """
+    Calculates the intersection over union.
+
+    Args:
+        preds: batch of prediction bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
+        target: batch of target bounding boxes with representation ``[x_min, y_min, x_max, y_max]``
```
akihironitta, line 10: May we have tensor sizes here as well?

akihironitta (changes requested):
@briankosw I left a suggestion. Would you mind considering it?

akihironitta:
> I'll pause work on this PR then.

Right, let's wait for what other members have to say about https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/469#issuecomment-748864389.

oke-aditya:
Also a small point, If we had to add say CIoU and DIoU (yeah IoUs are so many)
We could potentially then have a combined function.

```
def iou(type="ciou", boxes1: Tensor, boxe2: Tensor):

```

pl_bolts/metrics/object_detection.py
briankosw: Done!

Borda:
@akihironitta mind re-review? 🐰

akihironitta (approved):
LGTM!

Borda (approved):

Issue #409, opened 2020-11-26
ci: Fix possible OOM error `Process completed with exit code 137`


akihironitta:
## 🐛 Bug
Seems `CI full testing / pytest (ubuntu-20.04, *, *)` particularly tend to fail with the error:
```
/home/runner/work/_temp/5ef79e81-ccef-44a4-91a6-610886c324a6.sh: line 2:  1855 Killed                  coverage run --source pl_bolts -m pytest pl_bolts tests --exitfirst -v --junitxml=junit/test-results-Linux-3.7-latest.xml
Error: Process completed with exit code 137.
```
**Example CI runs**
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/runs/1459479942
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/runs/1459753659
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/runs/1459754977

This error might happen on different os or different versions. Haven't investigated yet.

### To Reproduce
Not sure how to reproduce...

### Additional context
Found while handling the dataset caching issue in https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/387#issuecomment-734396787.

Borda:
is it due to timeout? how long does it run before kill?

akihironitta:
The two runs were killed after 7m 43s and 7m 20s, so I guess it's not due to timeout...

Borda:
> The two runs were killed after 7m 43s and 7m 20s, so I guess it's not due to timeout...

you are right, the CI timeout is 45min so it can be some random failer or it is always the same test configuration?

akihironitta:
`CI full testing / pytest (ubuntu-20.04, 3.8, latest)` also fails in:
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/403/checks?check_run_id=1463269846.

Seems it happens particularly on Ubuntu...(?)

akihironitta:
It seems the following runs on Ubuntu were also unexpectedly killed due to probably the same reason:
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/400/checks?check_run_id=1463378727
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/400/checks?check_run_id=1463378796
- https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/400/checks?check_run_id=1463378820

[...]

PR #525, opened 2021-01-18
fixes DQN run_n_episodes using the wrong environment variable 

### LINKED ISSUE #516: DQN run_n_episodes ignores environment parameter
## 🐛 Bug

The method ``run_n_episodes`` takes an environment as parameter, that should be used to run a specific number of episodes. But instead it uses the objects environment (`self.env` and not the parameter `env`) to run the simulation steps.

This bug also causes warnings when testing the DQN model, as the passed environment is resetted, but the environment to run the simulation steps on is not reseted and thus may be in an invalid state (e.g. already finished).

### Additional context
The bug is caused by [line 175 in dqn_model.py](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/413b9df0d615fb6dbfe0bdc00f905ab7c4aed07d/pl_bolts/models/rl/dqn_model.py#L175).

sid-sundrani:
## What does this PR do?
DQN's `run_n_episodes` method, which is used by `test_step` for testing the agent, uses the object's environment `self.env` and not its argument `env` to run the simulation steps. It resets `env` but takes simulation steps in `self.env`. 

As a result, the testing stats are wrong because the wrong environment is used by the agent and a different environment is being reset after each episode

Fixes #516 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pl_bolts/models/rl/dqn_model.py
+++ b/pl_bolts/models/rl/dqn_model.py
@@ -171,7 +171,7 @@ class DQN(pl.LightningModule):
             while not done:
                 self.agent.epsilon = epsilon
                 action = self.agent(episode_state, self.device)
-                next_state, reward, done, _ = self.env.step(action[0])
+                next_state, reward, done, _ = env.step(action[0])
                 episode_state = next_state
                 episode_reward += reward
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the end of batch size mismatch ([#389](https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/389))
 - Fixed `batch_size` parameter for DataModules remaining ([#344](https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/344))
 - Fixed CIFAR `num_samples` ([#432](https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/432))
+- Fixed DQN `run_n_episodes` using the wrong environment variable ([#525](https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/525))
 
 ## [0.2.5] - 2020-10-12
 


pl_bolts/models/rl/dqn_model.py
```diff
@@ -171,7 +171,7 @@ def run_n_episodes(self, env, n_epsiodes: int = 1, epsilon: float = 1.0) -> List
             while not done:
                 self.agent.epsilon = epsilon
                 action = self.agent(episode_state, self.device)
-                next_state, reward, done, _ = self.env.step(action[0])
+                next_state, reward, done, _ = env.step(action[0])
```
Borda, line 174: shall we also assign the env back, `self.env = env`?

Borda (approved):

Borda:
it seems that the macOS test is hanging, mind have look at it?

pl_bolts/models/rl/dqn_model.py
sid-sundrani: if we assign the test env to `self.env` in `test_step`, when training is done after testing, it'll use the test env (without random seed) instead of what it was initialized with for training (env without random seed)

sid-sundrani:
> @sid-sundrani it seems that the macOS test is hanging, mind have look at it?

happening on 3.8. sure I'm looking into it

Borda:
> > @sid-sundrani it seems that the macOS test is hanging, mind have look at it?
> 
> happening on 3.8. sure I'm looking into it

it seems that it is hanging on master so most likely unrelated to this PR, but very welcome to find the hanging cases...

sid-sundrani:
> > > @sid-sundrani it seems that the macOS test is hanging, mind have look at it?
> > 
> > 
> > happening on 3.8. sure I'm looking into it
> 
> it seems that it is hanging on master so most likely unrelated to this PR, but very welcome to find the hanging cases...

yes, likely unrelated as the `TestValueAgent` tests pass

the hanging cases are unusual- in this PR, the pytest gets canceled after test_simclr, 3 more tests later than where the same test was canceled in the previous pr that was merged. i remember there was an issue in the past with perhaps OOM in a self_supervised failing test (#409)

Issue #200, opened 2020-09-10
Rethink usage pattern for pretrained models


nateraw:
## 🚀 Feature

Switch to using `SomeModel.from_pretrained('pretrained-model-name')` for pretrained models.

### Motivation

Seems we are following torchvision's pattern of having a 'pretrained' argument in the init of our models to initialize a pretrained model. In my opinion, this is extremely confusing as it makes the other init args + kwargs ambiguous/useless.

### Pitch

add `.from_pretrained` classmethod to models and initialize an instance of the class based off of that. Pretrained models should incorporate any hparams needed to fill out init, I guess.

```python
from pl_bolts.models import VAE

model = VAE.from_pretrained('imagenet2012')
```

williamFalcon:
yeah, agree... although this is basically just the same as load_from_checkpoint no? sounds like we're looking for checkpoint nicknames instead?

doesn't it read better as:

```
VAE.pretrained_on('xyz')
```

nateraw:
Right, I think the distinction here is that `load_from_checkpoint` is for checkpoints you have saved locally, but this function would be for pretrained models that we are hosting (i.e.[ these guys](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/850db3df9c6a2f0ae5afd32b921e046d556c54f6/pl_bolts/utils/pretrained_weights.py#L3-L11)).

So, yes! We are looking for something that can point to a nickname/identifier for a pretrained model.

---

I think 'pretrained_on' is a limiting name, as a model could be pretrained on the same dataset twice w/ different settings, and then would be ambiguous to load if using that function name. Thats why I suggest something a little more open, such as `from_pretrained(identifier)`. 

This is just my opinion... I could be convinced otherwise haha 😄 . Let's have others weigh in to come to consensus. 

CC: @PyTorchLightning/core-contributors

williamFalcon:
oh i see. it's an id not a dataset.
yeah that works.

for instance we can have many backbones with different datasets as well

```
CPC.from_pretrained('resnet18-imagenet')
CPC.from_pretrained('resnet50-imagenet')
CPC.from_pretrained('resnet18-stl10')
```

Borda:
Yes, they are trained on a defined dataset, in this case, the dataset name serves just as Look-up-table to a specific path on PL side...

ananyahjha93:
[...]

PR #529, opened 2021-01-19
Add RetinaNet Object detection with Backbones

### LINKED ISSUE #391: Add RetinaNet Object Detection
## 🚀 Feature

With the latest torchvision 0.8 release. RetinaNet is stable and we can use it.

### Motivation

Implement something similar to FRCNN. But after FRCNN is stable.

### Pitch

Something like #382 if it gets merged.

### Alternatives

I don't think so reimplementing RetinaNet would be wise.

### Additional context

I can probably add this after Faster RCNN is merged and is stable.

oke-aditya:
## What does this PR do?

Fixes #391 

## Before submitting

## PR review
 - [x] Is this pull request ready for review?

## Did you have fun?
I think yes :stuck_out_tongue:


oke-aditya:
This PR also bumps torchvision to 0.8.1 and PyTorch to 1.7.

pl_bolts/models/detection/faster_rcnn/faster_rcnn_module.py
```diff
@@ -117,7 +103,7 @@ def validation_step(self, batch, batch_idx):
         images, targets = batch
         # fasterrcnn takes only images for eval() mode
         outs = self.model(images)
-        iou = torch.stack([_evaluate_iou(t, o) for t, o in zip(targets, outs)]).mean()
+        iou = torch.stack([_evaluate_iou(o, t) for t, o in zip(targets, outs)]).mean()
```
oke-aditya, line 106: To keep consistency with `evaluate_iou`

pl_bolts/models/detection/retinanet/backbones.py
```diff
@@ -0,0 +1,41 @@
...
+from pl_bolts.utils import _TORCHVISION_AVAILABLE
+from pl_bolts.utils.warnings import warn_missing_pkg
+
+if _TORCHVISION_AVAILABLE:
+    from torchvision.models.detection.backbone_utils import resnet_fpn_backbone
+else:  # pragma: no cover
+    warn_missing_pkg("torchvision")
+
+
+def create_retinanet_backbone(
```
oke-aditya, line 15: This does code duplication.
There are Pros and Cons of both sides.

tests/models/test_detection.py
```diff
@@ -18,7 +18,7 @@ def test_fasterrcnn():
 
 
 def test_fasterrcnn_train(tmpdir):
-    model = FasterRCNN()
+    model = FasterRCNN(pretrained=False)
```
oke-aditya, line 21: I changed all pretrained to False. As CI coverage is not affected by these. It also speeds up by avoiding us to download the weights.

oke-aditya (commented):
I left some comments to ease further discussions :)

oke-aditya:
The RetinaNet test is very slow, the reason being not using Batched NMS implementation. This is fixed in torchvision master and we can expect faster RetinaNet model + tests in next torchvision release

pl_bolts/models/detection/__init__.py
```diff
@@ -1,7 +1,5 @@
 from pl_bolts.models.detection import components  # noqa: F401
 from pl_bolts.models.detection.faster_rcnn import FasterRCNN  # noqa: F401
+from pl_bolts.models.detection.retinanet import RetinaNet  # noqa: F401
```
akihironitta, line 3: We don't need these ignores?
```suggestion
from pl_bolts.models.detection import components
from pl_bolts.models.detection.faster_rcnn import FasterRCNN
from pl_bolts.models.detection.retinanet import RetinaNet
```

pl_bolts/models/detection/faster_rcnn/faster_rcnn_module.py
```diff
@@ -47,7 +33,7 @@ class FasterRCNN(pl.LightningModule):
     CLI command::
 
         # PascalVOC
-        python faster_rcnn.py --gpus 1 --pretrained True
+        python faster_rcnn_module.py --gpus 1 --pretrained True
```
akihironitta, line 36: Nice catch

tests/models/test_detection.py
```diff
@@ -18,19 +19,54 @@ def test_fasterrcnn():
...
 
     train_dl = DataLoader(DummyDetectionDataset(), collate_fn=_collate_fn)
     valid_dl = DataLoader(DummyDetectionDataset(), collate_fn=_collate_fn)
 
-    trainer = pl.Trainer(fast_dev_run=True, default_root_dir=tmpdir)
+    trainer = pl.Trainer(
+        fast_dev_run=True, logger=False, checkpoint_callback=False, max_epochs=1, default_root_dir=tmpdir
+    )
```
akihironitta, line 29: I guess we can drop `max_epochs=1`.

oke-aditya: Defensive programming in tests. :sweat_smile:

oke-aditya: Let me revert is :smile:

oke-aditya:
@Borda @akihironitta can you review 😅

pl_bolts/models/detection/retinanet/backbones.py
akihironitta: This is exactly the same as the one in FasterRCNN, right? Why don't we define both of `create_retinanet_backbone` and `create_fasterrcnn_backbone` in somewhere in `pl_bolts/models/detection/components/` and use it in both RetinaNet and FasterRCNN?
https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/4e1c1502b70e4be59a9ac80878ec5ad5a212f87e/pl_bolts/models/detection/faster_rcnn/backbones.py#L15

akihironitta (commented):
It'd be nice to hear your thoughts!

oke-aditya:
> This is exactly the same as the one in FasterRCNN, right? Why don't we define both of create_retinanet_backbone and create_fasterrcnn_backbone in somewhere in pl_bolts/models/detection/components/ and use it in both RetinaNet and FasterRCNN?


Wel yes, this can be refactored later.
Let me list my thoughts for pors and cons.

Pros
- As number of detection models increase it is better to keep them independent
- Maybe users are using our components and want to only create a model and train in their own way. This can be useful.
- Decoupling the code is better since users will find it natural to do `detection.create_detection_model_backone`
- We can patch a few differences between the models and it becomes bit easier to maintain each model and test it indepedently.

Cons: -
- Rewriting code. But I guess this will happen only for torchvision models.
- More tests. Well obviously, more code is more test.

Let me know your thoughts as of now we can leave it for the same.

oke-aditya:
CI Test Failure unreleated cc @akihironitta

akihironitta:
> CI Test Failure unreleated cc @akihironitta

You're right. #542 should fix the mypy error.

akihironitta:
@oke-aditya I agree on keeping models independent, but still, I think we should share reusable components to reduce redundancy where possible.

oke-aditya:
Lets keep it as it as for now. It's very little code duplication
This refactor is very simple and doesnt harm much currently.

oke-aditya:
@Borda  @akihironitta  can you review once :smiley:  This looks good to go from me

akihironitta (approved):

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,148 @@
...
+from pl_bolts.utils.warnings import warn_missing_pkg
+
+if _TORCHVISION_AVAILABLE:
+    from torchvision.models.detection.retinanet import RetinaNet as torchvision_RetinaNet
+    from torchvision.models.detection.retinanet import retinanet_resnet50_fpn, RetinaNetHead
+else:  # pragma: no cover
+    warn_missing_pkg("torchvision")
+
+
+class RetinaNet(pl.LightningModule):
```
Borda, line 19:
```suggestion
class RetinaNet(LightningModule):
```

pl_bolts/models/detection/faster_rcnn/faster_rcnn_module.py
Borda: why you have args `o, t` but zip `targets, outs` instead of `outs, targets`
the also rather call out as predits

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,148 @@
...
+            backbone: Pretained backbone CNN architecture.
+            fpn: If True, creates a Feature Pyramind Network on top of Resnet based CNNs.
+            pretrained: if true, returns a model pre-trained on COCO train2017
+            pretrained_backbone: if true, returns a model with backbone pre-trained on Imagenet
+            trainable_backbone_layers: number of trainable resnet layers starting from final block
+        """
+        super().__init__()
+        self.learning_rate = learning_rate
+        self.num_classes = num_classes
+        self.backbone = backbone
```
Borda, line 63: why not `save_hyperparameters()`?

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def training_step(self, batch, batch_idx):
+
+        images, targets = batch
+        targets = [{k: v for k, v in t.items()} for t in targets]
+
+        # fasterrcnn takes both images and targets for training, returns
+        loss_dict = self.model(images, targets)
+        loss = sum(loss for loss in loss_dict.values())
+        return {"loss": loss, "log": loss_dict}
```
Borda, line 92: use `self.log(...)`

requirements.txt
```diff
@@ -1,2 +1,2 @@
-torch>=1.6
+torch>=1.7
```
Borda, line 1: shy this is needed?

oke-aditya: Yes, torchvision 0.8 needs pytorch 1.7, so we need to bump :)

pl_bolts/models/detection/retinanet/retainanet_module.py
oke-aditya: Since I shifted the metrics, the args have changed.

I'm not sure how `save_hyperparameters()` works.

pl_bolts/models/detection/retinanet/retainanet_module.py
oke-aditya: Any example of `sef.log() ?`
I kept this consistent with FRCNN.

requirements.txt
Borda: ok, then make Retina as optional as we have with some others

pl_bolts/models/detection/retinanet/retainanet_module.py
Borda: yes but we shall be content with PL api
see for example https://pytorch-lightning.readthedocs.io/en/latest/logging.html#progress-bar

requirements.txt
oke-aditya: Ok we can make it optional, but why not update to PyTorch 1.7 ? It is out for 2 months and is stable. + CI didn't report any failure with PL 1.11 and torch 1.7 ?

Borda: yes but we do not force users to upgrade their env unless it is critical for all package

pl_bolts/models/detection/retinanet/retainanet_module.py
oke-aditya: Done. I hope it is correct

oke-aditya: Done. I hope it is correct

requirements.txt
oke-aditya: I'm not sure how to arrange this or refactor. @akihironitta @Borda   can you suggest changes / directly commit on top of this code ?

oke-aditya: I'm not sure how to arrange this or refactor. @akihironitta @Borda   can you suggest changes / directly commit on top of this code ?

akihironitta: @oke-aditya By setting `torchvision>=0.8` in [requirements/models.txt](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/requirements/models.txt), `pip install "pytorch-lightning-bolts[extra]"` will automatically install `torch>=1.7`, right? In that way, I think we can keep `torch>=1.6` as the basic requirement.

pl_bolts/models/detection/retinanet/retainanet_module.py
akihironitta: @oke-aditya Hope this example helps:
```python
import pytorch_lightning as pl

class Model(pl.LightningModule):
    def __init__(self, a, b, c):
        super().__init__()
        self.save_hyperparameters()
    def print(self):
        print(self.hparams, type(self.hparams))
        print(self.hparams.a, self.hparams.b, self.hparams.c)

model = Model(1, 2, 3)
model.print()
```
Output:
```
"a": 1
"b": 2
"c": 3 <class 'pytorch_lightning.utilities.parsing.AttributeDict'>
1 2 3
```

requirements.txt
akihironitta: So, shall we undo this change here and add `torchvision>=0.8` in requirements/models.txt?

pl_bolts/models/detection/faster_rcnn/faster_rcnn_module.py
akihironitta: @oke-aditya mind following the @Borda's comment?

oke-aditya: Tbh either doesn't matter. IoU is commutative.
Also let's merge this and refactor the metrics to use torchvision in a new PR.

pl_bolts/models/detection/retinanet/backbones.py
oke-aditya: Same thoughts, I think it will be fine to refactor later. Though, I feel it keeps a little bit code clarity and keeps little code independency.

requirements.txt
oke-aditya: Done !

oke-aditya:
Hmm this is because pytorch on `requirments.txt` is 1.6 and torchvision as 0.8 which is incompatible. Any solutions to this @akihironitta

kaushikb11:
Hi @oke-aditya, What are the blockers here?

oke-aditya:
Hi @kaushikb11 no potential blockers here, implementation is done, tests are added.

The blocker is that RetinaNet requires torchvision >= 0.8.1 and PyTorch >= 1,7
I proposed some solutions in Issue #554 . A simple fix to add torchvision >= 0.8.1 to `requirements/models.txt` does not work out well, it conflicts with PyTorch 1.6. 
Somehow we need to solve this torchvision dependency with CI testing too,

oke-aditya:
Thanks a lot @Borda for updating. Sorry that I didn't do much work

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,150 @@
...
+            fpn: If True, creates a Feature Pyramind Network on top of Resnet based CNNs.
+            pretrained: if true, returns a model pre-trained on COCO train2017
+            pretrained_backbone: if true, returns a model with backbone pre-trained on Imagenet
+            trainable_backbone_layers: number of trainable resnet layers starting from final block
+        """
+        super().__init__()
+        self.learning_rate = learning_rate
+        self.num_classes = num_classes
+        self.backbone = backbone
+        if backbone is None:
```
senarvi, line 64: This complexity of different combinations of arguments in the Faster R-CNN and RetinaNet models is quite confusing to me. If `backbone` ("*pretrained backbone CNN architecture*") is not given, one of the predefined Torchvision model architectures is used (`fasterrcnn_resnet50_fpn` or `retinanet_resnet50_fpn`), pretrained model and backbone may still be used, and `trainable_backbone_layers` is ignored. If `backbone` is given, then the `pretrained` argument is ignored.

We could improve the documentation to try and make this clear, but I was thinking that would it be more intuitive to
* make `backbone` a required argument in these models
* remove `pretrained` argument from these models
* add a third model `TorchvisionDetectionModel` that takes as an argument one of the predefined torchvision model names, `fasterrcnn_resnet50_fpn`, `fasterrcnn_mobilenet_v3_large_320_fpn`, `fasterrcnn_mobilenet_v3_large_fpn`, `retinanet_resnet50_fpn`, `keypointrcnn_resnet50_fpn`, `maskrcnn_resnet50_fpn`, etc. and supports also loading an entirely pretrained model.

oke-aditya: Backbones have a problem. Not all backbones work with FPN.
Not all have pretrained weights on COCO available, some have on imagenet. 
Pretrained argument is needed, so we can't remove it.
Pretrainde either load COCO weights or ImageNet weights.

This can be made explicit making pretrained an 'str' which can be = 'imagenet' or 'coco')
But this was deferred for #200 .

Current API works as follows.
The reason for `pretraind_backbone_layers` being ingored is [here](https://github.com/pytorch/vision/issues/3406)
Since it is fixed, we can pass that param. But again this needs a mandatory torchvision bump which cannot be enforced.
See #554 .

Current API works as

```
model = RetinaNet(backbone="resnet50", fpn=True)
```

We can instead refactor as follows: -
All detection models accept backbone as `nn.Module` not an `str`.
Users should create a backbone using the method provided to `create_xyz_backbone`.

E.g.  We have provided `create_retinanet_backbone`

```


backbone_model = retinanet.create_retinanet_backbone(backbone, fpn, pretrained_backbone, **kwargs)

Here backbone is "str", so user can pass "resnet50", "mobilenet" etc.
We can easily load all the above supported models.

model = RetinaNet(backbone_model, num_classes, learning_rate, **kwargs)

```

This separates both backbone creation and model creation, plus offers a clear API for all models.

To directly access pre trained models, users can simply create them.

`TorchvisionDetectionModel` is a fine abstraction, but do we need to complicate it ?
User would like to subclass RetinaNet and continue his customization, hence I feel to avoid this.

Also cc @kaushikb11 since flash too has a similar scenario to handle backbones.

senarvi: `pretrained` argument is needed only when `backbone` is not given. What I meant is that these models would work as if `backbone` is always given (and pretrained models cannot be accessed), whereas `TorchvisionDetectionModel` could load pretrained models.

I like your suggestion of separating backbone and model creation. The user could explicitly create a pretrained backbone. But I don't see how the user could access pretrained model weights.

oke-aditya: I'm not sure if I understood `pretrained` problem.
But this handles as follows (similar to torchvision's handling of `pretrained_backbone` and `pretrained` params)

For backbones, the pretrained weights could be loaded from imagenet if pretrained is passed `true` there.

And if the user passes pretrained in model `RetinaNet` we will try to load `COCO` weights.

Passing both true would not work (it does not in torchvision too you cannot pass `pretrained_backbone` and `pretrained` as True)

E.g. to load ImageNet weights and create pretrained FRCNN / RetinaNet

```
backbone = retinanet.create_retinanet_backbone(backbone="resnet50", fpn, pretrained_backbone=True,,  **kwargs)
model = RetinaNet(backbone, learning_rate, pretrained=False, **kwargs)
```

To load pretrained COCO weights (if available, very few combinations have it)

```
backbone = retinanet.create_backbone(backbone="resnet50", fpn=True, pretrained_backbone=False)
model = RetinaNet(backbone, pretrained=True, num_classes)
```

Note that COCO weights are only available in 3 combinations
`resnet50_fpn`, `mobilnet_large_fpn`, `mobilenet_320_fpn`.

For other cases, the user should initialize with imagenet pretrained backbone (preferable).

Yes, backbone is of course always necessary to create a model. But if we split up backbone creation and model creation it would be easier.

Currently the user has to pass a lot of params to a single Model Definition, which is creating confusion.

senarvi: In your proposal, the backbone argument can be any `nn.Module`, right? I thought it might be tricky then to know which weights to load, if the user creates RetinaNet with `pretrained=True`. But it sounds like you've thought it out. That would indeed be the most intuitive interface!

oke-aditya: Great. Nice to discuss and comeup with solution! Great ideas @senarvi. 
I will once request @akihironitta  @Borda  to confirm. I will refactor the code accordingly!

Also cc @kaushikb11  if flash has better solutions / ideas or anything if we missed.

Borda:
@oke-aditya could you please check the conflits?

oke-aditya:
I will have a look this week

oke-aditya:
Sorry for the delay. I think I will find time this week. I think conflicts can easily be resolved, I did have a look with github editor.
Is there any other work to be done in this PR?

Borda:
> I'm bumping the torch and torchvision for CI, This is the minimum version needed for retinanet.

<I think it is fine, just pls have look at conflict so we can land it...

oke-aditya:
Conflicts resolved. Should be good to go :rocket:

oke-aditya:
Any idea to fix the CI error. It is possibly due to incorrect IoU implementation / due to model not detecting any boxes.

Borda:
@SkafteNicki mind have look? :]

pl_bolts/models/detection/faster_rcnn/faster_rcnn_module.py
```diff
@@ -107,14 +98,14 @@ def training_step(self, batch, batch_idx):
...
-        return {"loss": loss, "log": loss_dict}
+        self.log("loss", loss, prog_bar=True)
 
     def validation_step(self, batch, batch_idx):
         images, targets = batch
         # fasterrcnn takes only images for eval() mode
         outs = self.model(images)
         iou = torch.stack([_evaluate_iou(t, o) for t, o in zip(targets, outs)]).mean()
-        return {"val_iou": iou}
+        self.log("val_iou", iou, prog_bar=True)
```
SkafteNicki, line 108: return statement is still needed for the computations in `validation_epoch_end`
```suggestion
        self.log("val_iou", iou, prog_bar=True)
        return {"val_iou2: iou}
```
should fix the tests

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,150 @@
...
+    def training_step(self, batch, batch_idx):
+
+        images, targets = batch
+        targets = [{k: v for k, v in t.items()} for t in targets]
+
+        # fasterrcnn takes both images and targets for training, returns
+        loss_dict = self.model(images, targets)
+        loss = sum(loss for loss in loss_dict.values())
+        self.log("loss", loss, prog_bar=True)
+
```
SkafteNicki, line 95: this should return the loss right?
```suggestion
        self.log("loss", loss, prog_bar=True)
        return loss

```

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,150 @@
...
+    def validation_step(self, batch, batch_idx):
+        images, targets = batch
+        # fasterrcnn takes only images for eval() mode
+        outs = self.model(images)
+        iou = torch.stack([_evaluate_iou(t, o) for t, o in zip(targets, outs)]).mean()
+        self.log("val_iou", iou, prog_bar=True)
+
```
SkafteNicki, line 102: return so they are available in `validation_epoch_end`
```suggestion
        self.log("val_iou", iou, prog_bar=True)
        return {"val_iou": iou}

```

pl_bolts/models/detection/faster_rcnn/faster_rcnn_module.py
```diff
@@ -107,13 +98,14 @@ def training_step(self, batch, batch_idx):
         # fasterrcnn takes both images and targets for training, returns
         loss_dict = self.model(images, targets)
         loss = sum(loss for loss in loss_dict.values())
-        return {"loss": loss, "log": loss_dict}
+        self.log("loss", loss, prog_bar=True)
```
SkafteNicki, line 101:
```suggestion
        self.log("loss", loss, prog_bar=True)
        return loss
```

oke-aditya:
Looks like CI went good. Hopefully it gets merged

SkafteNicki (approved):

requirements.txt
```diff
@@ -1,4 +1,4 @@
-torch>=1.6
+torch>=1.7.1
```
SkafteNicki, line 1: Currently GPU tests are failing with:
```E   OSError: /usr/local/lib/python3.7/dist-packages/torchtext/_torchtext.so: undefined symbol: _ZNK3c104Type14isSubtypeOfExtESt10shared_ptrIS0_EPSo```
I assume it has something to do with the update to the torch version

Comment (oke-aditya) [reply requirements.txt:1]: I'm unsure, but we have to bump pytorch and torchvision to 1.7.1 and 0.8.2 to use RetinaNet

**Comment by Borda** [comment]
@oke-aditya could you pls resolve conflict, then we shall be fine to go... :rabbit:

**Comment by oke-aditya** [comment]
There aren't any conflicts @Borda  I guess this is good to go :rabbit:

**Comment by Borda** [comment]
> There aren't any conflicts @Borda I guess this is good to go rabbit

yeah, I have fixed them, lets see the CI finish

**Comment by oke-aditya** [comment]
Thnx for the fix! I didn't notice that

**Comment by Borda** [comment]
@oke-aditya mind check the last failing tests?

**Comment by oke-aditya** [comment]
CI Failures Look unrelated to the code change
```

(791 durations < 0.005s hidden.  Use -vv to show these durations.)
=========================== short test summary info ============================
FAILED tests/models/test_scripts.py::test_cli_run_srgan[--dataset mnist --scale_factor 4 --data_dir /home/runner/work/lightning-bolts/lightning-bolts/datasets --max_epochs 1 --max_steps 2 --batch_size 4]
FAILED tests/models/test_scripts.py::test_cli_run_srresnet[--dataset mnist --scale_factor 4 --data_dir /home/runner/work/lightning-bolts/lightning-bolts/datasets --max_epochs 1 --max_steps 2 --batch_size 4]
```

```
test_cli_run_srgan[--dataset mnist --scale_factor 4 --data_dir /Users/runner/work/lightning-bolts/lightning-bolts/datasets --max_epochs 1 --max_steps 2 --batch_size 4] _

cli_args = '--dataset mnist --scale_factor 4 --data_dir /Users/runner/work/lightning-bolts/lightning-bolts/datasets --max_epochs 1 --max_steps 2 --batch_size 4'

    @pytest.mark.parametrize("cli_args", ["--dataset mnist --scale_factor 4" + _DEFAULT_ARGS])
    def test_cli_run_srgan(cli_args):
        from pl_bolts.models.gans.srgan.srgan_module import cli_main
    
        with mock.patch("argparse._sys.argv", ["any.py"] + cli_args.strip().split()):
>           cli_main()

tests/models/test_scripts.py:38: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pl_bolts/models/gans/srgan/srgan_module.py:194: in cli_main
    parser = TVTDataModule.add_argparse_args(parser)
../../../hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/pytorch_lightning/core/datamodule.py:241: in add_argparse_args
    return add_argparse_args(cls, parent_parser, **kwargs)
../../../hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/pytorch_lightning/utilities/argparse.py:207: in add_argparse_args
    args_and_types = get_init_arguments_and_types(symbol)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <function TVTDataModule.__init__ at 0x13fd18f28>

    def get_init_arguments_and_types(cls) -> List[Tuple[str, Tuple, Any]]:
        r"""Scans the class signature and returns argument names, types and default values.
    
        Returns:
            List with tuples of 3 values:
            (argument name, set with argument types, argument default value).
    
        Examples:
    
            >>> from pytorch_lightning import Trainer
            >>> args = get_init_arguments_and_types(Trainer)
    
        """
        cls_default_params = inspect.signature(cls).parameters
        name_type_default = []
        for arg in cls_default_params:
            arg_type = cls_default_params[arg].annotation
            arg_default = cls_default_params[arg].default
            try:
>               arg_types = tuple(arg_type.__args__)
E               TypeError: 'NoneType' object is not iterable

../../../hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/pytorch_lightning/utilities/argparse.py:128: TypeError
----------------------------- Captured stderr call -----------------------------
Global seed set to 1234
_ test_cli_run_srresnet[--dataset mnist --scale_factor 4 --data_dir /Users/runner/work/lightning-bolts/lightning-bolts/datasets --max_epochs 1 --max_steps 2 --batch_size 4] _

cli_args = '--dataset mnist --scale_factor 4 --data_dir /Users/runner/work/lightning-bolts/lightning-bolts/datasets --max_epochs 1 --max_steps 2 --batch_size 4'
```

They come from GANs IMO

**Comment by oke-aditya** [comment]
As well as Mypy

```
pl_bolts/callbacks/ssl_online.py:58: error: Attribute "num_classes" already defined on line 56  [no-redef]
pl_bolts/callbacks/ssl_online.py:59: error: Attribute "dataset" already defined on line 57  [no-redef]
pl_bolts/callbacks/ssl_online.py:98: error: Item "None" of "Optional[SSLEvaluator]" has no attribute "parameters"  [union-attr]
pl_bolts/callbacks/ssl_online.py:101: error: Item "None" of "Optional[SSLEvaluator]" has no attribute "load_state_dict"  [union-attr]
pl_bolts/callbacks/ssl_online.py:119: error: Function is missing a return type annotation  [no-untyped-def]
pl_bolts/callbacks/ssl_online.py:130: error: unused "type: ignore" comment
pl_bolts/callbacks/ssl_online.py:130: error: "None" not callable  [misc]
pl_bolts/callbacks/ssl_online.py:150: error: Item "None" of "Optional[Any]" has no attribute "step"  [union-attr]
pl_bolts/callbacks/ssl_online.py:151: error: Item "None" of "Optional[Any]" has no attribute "zero_grad"  [union-attr]
pl_bolts/callbacks/ssl_online.py:170: error: Item "None" of "Optional[SSLEvaluator]" has no attribute "state_dict"  [union-attr]
pl_bolts/callbacks/ssl_online.py:170: error: Item "None" of "Optional[Any]" has no attribute "state_dict"  [union-attr]
pl_bolts/callbacks/ssl_online.py:177: error: Function is missing a return type annotation  [no-untyped-def]
```

I guess nothing to do from this PR perspective

cc  @Borda

**Comment by oke-aditya** [comment]
By the way, what is the future of bolts? Or is it recommended users should migrate to Flash?

**Comment by Borda** [comment]
> By the way, what is the future of bolts? Or is it recommended users should migrate to Flash?

that is a good question... @ethanwharris?
 let's sync on Slack or have a call next week...

**Comment by oke-aditya** [comment]
I would be interested to know! 
Note that torchvision is getting a mega rewrite to support lot of new features. So probably something to note too.

File: pl_bolts/metrics/object_detection.py [hunk 17]
```diff
@@ -2,7 +2,16 @@
...
+def _evaluate_iou(preds: torch.Tensor, target: torch.Tensor):
+    """Evaluate intersection over union (IOU) for target from dataset and output prediction from model."""
+
+    if preds["boxes"].shape[0] == 0:
+        # no box detected, 0 IOU
+        return torch.tensor(0.0, device=preds["boxes"].device)
+    return iou(target["boxes"], preds["boxes"]).diag().mean()
+
+
+def iou(preds: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
```
Borda, line 14: can we use rather a TorchMetrics version?

pl_bolts/models/detection/retinanet/retainanet_module.py
```diff
@@ -0,0 +1,152 @@
...
+    def configure_optimizers(self):
+        return torch.optim.SGD(
+            self.model.parameters(),
+            lr=self.learning_rate,
+            momentum=0.9,
+            weight_decay=0.005,
+        )
+
+    @staticmethod
+    def add_model_specific_args(parent_parser):
```
Borda, line 119: can we use Lightning CLI instead?

Borda (commented):
looks good just minor suggestions...

Borda:
@akihironitta could you pls have look at this one so we can include it in the release at beginning of next week...

pl_bolts/metrics/object_detection.py
oke-aditya: Sorry for the delay. I wasn't well past week.
I could :( But it would take more time for me as I haven't used torchmetrics up lately.

akihironitta: I don't think we have a iou metric for detection in torchmetrics, so let's keep as is for now.

oke-aditya:
Thanks a lot @akihironitta Akki!

Sorry from my side for not working on this.

akihironitta:
The currently failing tests are irrelevant to this PR. The failures are due to https://github.com/PyTorchLightning/pytorch-lightning/issues/9535 where the error occurs only with Python 3.6 and `pytorch_lightning<1.4.8`. The fix (https://github.com/PyTorchLightning/pytorch-lightning/pull/9554) is included in `>=1.4.8`.

pl_bolts/metrics/object_detection.py
akihironitta: FYI, RetinaNet now uses `torchvision.ops.box_iou` instead of our in-house `iou` here.

akihironitta (approved):
I believe this PR is ready for getting merged.

Just for the note, here's the summary of my last tens of commits:

- Updated CHANGELOG
- Added its doc
- Reverted the change in Faster R-CNN as it's irrelevant to this PR
- Made RetinaNet use `torchvision.ops.box_iou` instead of our in-house `pl_bolts.metrics.iou`
- Resolved all the failing tests related to this PR

cc @Borda

requirements.txt
```diff
@@ -1,5 +1,5 @@
...
 torchmetrics>=0.4.1
-pytorch-lightning>=1.3.0
+pytorch-lightning>=1.4.8
```
Borda, line 3: @akihironitta just wondering why we need here so high PL version?

Borda:
> Just for the note, the currently failing tests are irrelevant to this PR. The failures are due to [PyTorchLightning/pytorch-lightning#9535](https://github.com/PyTorchLightning/pytorch-lightning/issues/9535) where the error occurs only with Python 3.6 and `pytorch_lightning<1.4.8`. The fix ([PyTorchLightning/pytorch-lightning#9554](https://github.com/PyTorchLightning/pytorch-lightning/pull/9554)) is included in `>=1.4.8`.

so if we set py3.7 as the min version then we do not need to upgrade PL?

akihironitta:
TODO:
- [ ] Set the min PL version to `1.4.0`
- [ ] Remove `argparse` and use `LightningCLI`

oke-aditya:
Looks great. So kind of you @akihironitta Aki : 💖

oke-aditya:
Thanks a lot to @akihironitta and @Borda for your patience. I'm Extremely sorry for this being delayed.

akihironitta:
@oke-aditya Hey Aditya, not at all! Thank you for your work and patience for almost a year :tada:

Issue #427, opened 2020-12-05
Tidy up the model subpackages


akihironitta:
## 🚀 Refactoring
The structure in `pl_bolts.models` is not user-friendly due to:

- an inconsistent structure, for example:
   ```python
   from pl_bolts.models.vision import SemSegment
   from pl_bolts.models.vision import ImageGPT    # This doesn't work,
   from pl_bolts.models import ImageGPT           # but this works.
   ```

- ~several ways to import a certain class, for example:~ <- Shall we discuss this in #423?
   ```python
   from pl_bolts.models import LogisticRegression
   from pl_bolts.models.regression import LogisticRegression
   from pl_bolts.models.regression.logistic_regression import LogisticRegression
   ```

### Pitch
Two options I have in mind to make it consistent are:

1. Make all models available right under `pl_bolts.models`
   ```python
   pl_bolts.models.VAE
   pl_bolts.models.AE
   pl_bolts.models.ImageGPT
   pl_bolts.models.PixelCNN
   pl_bolts.models.SwAV
   pl_bolts.models.SimCLR
   ...
   ```

2. Split models into domains, e.g. `vision`, `self_supervised`... (which I prefer)
   ```python
   pl_bolts.models.autoencoders.VAE
   pl_bolts.models.autoencoders.AE
   pl_bolts.models.vision.ImageGPT
   pl_bolts.models.vision.PixelCNN
   pl_bolts.models.self_supervised.SwAV
   pl_bolts.models.self_supervised.SimCLR
   ...
   ```

@Borda @ananyahjha93 @annikabrundyn @oke-aditya I'd love to hear your opinions on if it should be as is or another interface :]

oke-aditya:
I feel 2nd is better. I also prefer that there should be only one interface to import a model / function.

So API number 1 `from pl_bolts.models import LogisticRegression` doesn't work.

Still users can do `from pl_bolts.models.regression.logistic_regression import LogisticRegression`.  

We cannot stop this any way but this is highly discouraged. This is nature of python, people are browsing to a particular file and importing a function. This allows access to `_foobar` functions as well. Even if we don't expose `__init__` the above is still possible. (though we should ensure we haven't exposed such modules in `__init__`).
So we cannot refactor to have only 1 way of importing, there will be these 2 (inevitable).

I am not sure if we should warn people if they do above, but the possibly users should do 
`from pl_bolts.models.regression import LogisticRegression` .

[...]

PR #552, opened 2021-02-02
Add YOLO object detection model

### LINKED ISSUE #22: Add yolo v3/4
https://github.com/eriklindernoren/PyTorch-YOLOv3/issues/487
https://github.com/Tianxiaomo/pytorch-YOLOv4

senarvi:
## What does this PR do?

This PR adds the YOLO object detection model. The implementation is based on the YOLOv3 and YOLOv4 Darknet implementations, although it doesn't include all the features of YOLOv4. Detection seems to work with weights that have been trained using the Darknet implementation, so the network architecture should be more or less identical. The network architecture is read from a configuration file in the same format as in the Darknet implementation. It supports loading weights from a Darknet model file too, if you don't want to start training from a randomly initialized model.

Fixes #22 

## Before submitting

## PR review

## Did you have fun?


pl_bolts/datamodules/vocdetection_datamodule.py
```diff
@@ -152,16 +152,20 @@ def prepare_data(self) -> None:
         VOCDetection(self.data_dir, year=self.year, image_set="val", download=True)
 
     def train_dataloader(
-        self, batch_size: int = 1, image_transforms: Union[List[Callable], Callable] = None
+        self,
+        batch_size: int = 1,
+        transforms: Optional[List[Callable]] = [],
```
akihironitta, line 157: I think we don't need `Optional` here if we set `[]` as the default value.
```suggestion
        transforms: List[Callable] = [],
```

akihironitta (commented):
We recently fixed some of the tests, so would you mind merging master branch? Also, it seems the change here is causing errors in CI, so would you mind resolving that, too?

pl_bolts/datamodules/vocdetection_datamodule.py
senarvi: > We recently fixed some of the tests, so would you mind merging master branch? Also, it seems the change here is causing errors in CI, so would you mind resolving that, too?

I fixed the datamodule, but what do you mean by merging master branch? I believe this is based on the head of the master branch.

senarvi: I'm resolving this conversation but let me know if there's still something I need to do.

williamFalcon:
this is awesome. @Borda @kaushikb11 let's get this merged? and can we add as backbones available in Flash?

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,553 @@
...
+    area2 = dims2[:, 0] * dims2[:, 1]  # [M]
+
+    inter_wh = torch.min(dims1[:, None, :], dims2)  # [N, M, 2]
+    inter = inter_wh[:, :, 0] * inter_wh[:, :, 1]  # [N, M]
+    union = area1[:, None] + area2 - inter  # [N, M]
+
+    return inter / union
+
+
+def _elementwise_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
```
oke-aditya, line 58: hmm, this is same as `torchvision.ops.box_iou` ? I guess ?

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,553 @@
...
+    rb = torch.min(boxes1[:, 2:], boxes2[:, 2:])  # [N,2]
+
+    wh = (rb - lt).clamp(min=0)  # [N,2]
+    inter = wh[:, 0] * wh[:, 1]  # [N]
+
+    iou = inter / (area1 + area2 - inter)
+    return iou
+
+
+def _elementwise_generalized_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
```
oke-aditya, line 85: We recently added this to torchvision 0.8 release `torchvision.ops.generalized_box_iou`

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,553 @@
...
+    lti = torch.min(boxes1[:, :2], boxes2[:, :2])
+    rbi = torch.max(boxes1[:, 2:], boxes2[:, 2:])
+
+    whi = (rbi - lti).clamp(min=0)  # [N,2]
+    areai = whi[:, 0] * whi[:, 1]
+
+    return iou - (areai - union) / areai
+
+
+class IoULoss(nn.Module):
```
oke-aditya, line 126: 1 - `box_iou` simply ?

oke-aditya (commented):
The PR is great :tada:  Nice implementation of YOLO v4. It's hard to do in PyTorch since it's model is originally written in DarkNet. :sweat_smile:  Great work


A few comments, the reason is [this](https://github.com/pytorch/vision/pull/3383) pr on torchvision. Hence I think it could be safer to use `torchvision.ops` ?

Somehow torchvision dependency continues here as well I proposed a few solutions in #554 .
We actually nativelely implemented ops in bolts #347 #370 . But either we need to update them cause of above issue or use torchvision implementation.

IIRC torchvision 0.7 did not document these functions for IoU, box_area, and they were not stable.
GIoU and Tests were added in 0.8 to make it generally available.

pl_bolts/models/detection/yolo/yolo_layers.py
senarvi: @oke-aditya It's basically the same thing, but performs the operation elementwise on two sets of N boxes, producing N results. The torchvision op performs the operation on every pair formed from two sets of boxes, producing an N*M matrix. In other words, this is the same as taking the diagonal of `torchvision.ops.box_iou`, however, that would take quadratic time. The most generic and consistent way would probably be that the torchvision `box_iou` and `generalized_box_iou` functions would perform the elementwise operation, and for creating a pairwise matrix one would add empty dimensions to the inputs (or include an optional argument in those functions to perform the elementwise operation).

pl_bolts/models/detection/yolo/yolo_layers.py
senarvi: See the discussion about `_elementwise_iou` above.

pl_bolts/models/detection/yolo/yolo_layers.py
senarvi: To be as generic as possible, any loss function can be given to the detection layer as a parameter. So something like this would be possible: `lambda inputs, target: 1.0 - _elementwise_iou(inputs, target)`. Not that `lambda inputs, target: 1.0 - box_iou(inputs, target)` is a different thing. It calculates the loss between every (detector, target) pair, while YOLO matches the detectors to the targets and only calculates the loss from the matched pairs only.

senarvi:
> The PR is great 🎉 Nice implementation of YOLO v4. It's hard to do in PyTorch since it's model is originally written in DarkNet. 😅 Great work
> 
> A few comments, the reason is [this](https://github.com/pytorch/vision/pull/3383) pr on torchvision. Hence I think it could be safer to use `torchvision.ops` ?

Thanks! Indeed it's somewhat problematic how this project overlaps with torchvision. I definitely agree that I would rather use torchvision functions than copy them here, for the exact reason that you pointed out. We could use `box_iou(images, target).diagonal()` if we think it's worth the performance hit.

pl_bolts/models/detection/yolo/__init__.py
```diff
@@ -0,0 +1,4 @@
+from pl_bolts.models.detection.yolo.yolo_config import YoloConfiguration
+from pl_bolts.models.detection.yolo.yolo_module import Yolo
+
+__all__ = ["YoloConfiguration", "Yolo"]
```
akihironitta, line 4: Could we make this file empty? Otherwise, we'll have two ways to import these names:
```python
from pl_bolts.models.detection import Yolo       # <- This is fine.
from pl_bolts.models.detection.yolo import Yolo  # <- We don't want this to keep the API simple. (I think)
```

akihironitta (commented):
@senarvi This is well documented and a great addition to Bolts! I just left some comments, so would you mind having a look?

pl_bolts/models/detection/yolo/yolo_layers.py
akihironitta: Same as https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/552#discussion_r575473956, right?

pl_bolts/models/detection/yolo/yolo_layers.py
akihironitta: So, here we have two options (right?):
1. to use `_elementwise_iou` as is, or
2. to remove `_elementwise_iou` and use `torchvision.ops.box_iou` (but increase computational cost from `O(N)` to `O(N^2)`)

@oke-aditya @PyTorchLightning/bolts-contributors @Borda Any thoughts on this? (I'm neutral on this.)

pl_bolts/models/detection/yolo/yolo_config.py
```diff
@@ -0,0 +1,265 @@
...
+import torch.nn as nn
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+
+from pl_bolts.models.detection.yolo import yolo_layers
+
+
+class YoloConfiguration:
+    """
+    This class can be used to parse the configuration files of the Darknet YOLOv4 implementation.
+    The `get_network()` method returns a PyTorch module list that can be used to construct a YOLO
```
akihironitta, line 14: Single quote will be rendered in italic in the docs i.e. *get_network()*, so we need double quotes here to make it look like `get_network()`.
```suggestion
    The ``get_network()`` method returns a PyTorch module list that can be used to construct a YOLO
```

pl_bolts/models/detection/yolo/__init__.py
oke-aditya: I think we can just rename yolo.py `yolo_module` like we have done for other detection models

pl_bolts/models/detection/yolo/yolo_layers.py
oke-aditya: It is fine to increase a slight computational cost. Since these are tensor operations, they are fast and batched, it will not scale up so soon.

senarvi: I might add a third option, suggesting a change in the torchvision interface that supports both variants. But it's better to merge this PR first and improve it in future PRs, so I'll do like @oke-aditya suggests and use the current torchvision interface for now. At least then we'll know that it works correctly.

pl_bolts/models/detection/yolo/__init__.py
senarvi: @oke-aditya there's no yolo.py. There's a yolo subdirectory, like faster_rcnn.  @akihironitta FasterRCNN is also imported in both [detection/__init__.py](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/pl_bolts/models/detection/__init__.py) and [detection/faster_rcnn/__init__.py](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/pl_bolts/models/detection/faster_rcnn/__init__.py). Should I clear `detection/faster_rcnn/__init__.py` too, or leave them like this?

oke-aditya: Nice catch. I will solve faster RCNN in Retina Net PR. This will avoid merge conflicts too. 
Hopefully I will get retina Net completed by this week. (Really sorry for lack of time)

pl_bolts/models/detection/yolo/yolo_module.py
```diff
@@ -0,0 +1,516 @@
...
+    from torchvision.ops import nms
+    from torchvision.transforms import functional as F
+except ModuleNotFoundError:
+    warn_missing_pkg('torchvision')  # pragma: no-cover
+    _TORCHVISION_AVAILABLE = False
+else:
+    _TORCHVISION_AVAILABLE = True
+
+
+class Yolo(pl.LightningModule):
```
kaushikb11, line 26: Nit: I prefer `YOLO` in caps. @akihironitta What do you think?

pl_bolts/models/detection/yolo/yolo_module.py
```diff
@@ -0,0 +1,516 @@
...
+        if targets is None:
+            return boxes, confidences, classprobs
+
+        losses = {loss_name: mean_loss(loss_name) for loss_name in losses[0].keys()}
+        return boxes, confidences, classprobs, losses
+
+    def configure_optimizers(self) -> Tuple[List, List]:
+        """Constructs the optimizer and learning rate scheduler."""
+        if self.optimizer == 'sgd':
+            optimizer = optim.SGD(
```
kaushikb11, line 172: What not pass `optimizer: Type[Optimizer] = torch.optim.Adam` instead of `str`? Wouldn't need the conditions here.

pl_bolts/models/detection/yolo/yolo_module.py
akihironitta: I agree. @Separius Would you mind following the @kaushikb11's suggestion above?

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,553 @@
...
+
+from pl_bolts.utils.warnings import warn_missing_pkg
+
+try:
+    from torchvision.ops import box_area, box_iou
+except ModuleNotFoundError:
+    warn_missing_pkg('torchvision')  # pragma: no-cover
+    _TORCHVISION_AVAILABLE = False
+else:
+    _TORCHVISION_AVAILABLE = True
```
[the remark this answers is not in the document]
senarvi: I'm now using `pl_bolts.utils._TORCHVISION_AVAILABLE` in a couple of places.

pl_bolts/models/detection/yolo/yolo_module.py
```diff
@@ -0,0 +1,516 @@
...
+
+try:
+    import torchvision.transforms as T
+    from torchvision.ops import nms
+    from torchvision.transforms import functional as F
+except ModuleNotFoundError:
+    warn_missing_pkg('torchvision')  # pragma: no-cover
+    _TORCHVISION_AVAILABLE = False
+else:
+    _TORCHVISION_AVAILABLE = True
```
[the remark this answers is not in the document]
senarvi: Fixed.

pl_bolts/models/detection/yolo/yolo_config.py
senarvi: I replaced single quotes with double quotes in several docstrings.

pl_bolts/models/detection/yolo/yolo_module.py
senarvi: I guess I wrote it this way because it was easy to add the constructor arguments to the CLI arguments using `argparse_utils.add_argparse_args()`, but you're right - it's a lot better to give the opportunity to provide any kind of optimizer object.

pl_bolts/models/detection/yolo/yolo_module.py
senarvi: Right, the convention seems to be to write acronyms with all caps. (By the way, my handle is @senarvi.)

pl_bolts/models/detection/yolo/yolo_config.py
```diff
@@ -0,0 +1,265 @@
...
+from typing import List, Tuple
+from warnings import warn
+
+import torch.nn as nn
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+
+from pl_bolts.models.detection.yolo import yolo_layers
+
+
+class YoloConfiguration:
```
kaushikb11, line 11:
```suggestion
class YOLOConfiguration:
```

pl_bolts/models/detection/yolo/__init__.py
akihironitta: > @akihironitta FasterRCNN is also imported in both [detection/__init__.py](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/pl_bolts/models/detection/__init__.py) and [detection/faster_rcnn/__init__.py](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/master/pl_bolts/models/detection/faster_rcnn/__init__.py). Should I clear `detection/faster_rcnn/__init__.py` too, or leave them like this?

@senarvi There are similar structures in `pl_bolts/models/*` as you point out, and we are actually trying to refactor these as discussed in https://github.com/PyTorchLightning/pytorch-lightning-bolts/issues/427#issuecomment-762032327, so let's make this file empty.

pl_bolts/models/detection/yolo/yolo_config.py
senarvi: Fixed.

senarvi: Fixed.

pl_bolts/models/detection/yolo/__init__.py
senarvi: Fixed.

pl_bolts/models/detection/yolo/yolo_layers.py
senarvi: I'm using torchvision now. By the way, the computational cost is quadratic in the _number of targets per image_, which is typically not very large. `generalized_box_iou` is not yet supported by the version of torchvision used by the PL Bolts test suite, so I added a check.

pl_bolts/datamodules/vocdetection_datamodule.py
```diff
@@ -176,15 +180,21 @@ def train_dataloader(
         )
         return loader
 
-    def val_dataloader(self, batch_size: int = 1, image_transforms: Optional[List[Callable]] = None) -> DataLoader:
+    def val_dataloader(
+        self,
+        batch_size: int = 1,
+        transforms: List[Callable] = [],
+        image_transforms: Optional[Callable] = None
```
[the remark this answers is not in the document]
senarvi: Actually it doesn't. [`vocdetection_datamodule.Compose` assumes that `image_transforms` is callable](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/3e9bddeb09a71fafc55769fa9a5c0cd7033cd330/pl_bolts/datamodules/vocdetection_datamodule.py#L30), so when normalization is used, [the default image transforms are wrapped in a `torchvision.transforms.Compose`](https://github.com/PyTorchLightning/pytorch-lightning-bolts/blob/3e9bddeb09a71fafc55769fa9a5c0cd7033cd330/pl_bolts/datamodules/vocdetection_datamodule.py#L214).

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,553 @@
...
+        num_classes: int,
+        image_width: int,
+        image_height: int,
+        anchor_dims: List[Tuple[int, int]],
+        anchor_ids: List[int],
+        xy_scale: float = 1.0,
+        ignore_threshold: float = 0.5,
+        overlap_loss_func: Callable = None,
+        class_loss_func: Callable = None,
+        confidence_loss_func: Callable = None,
```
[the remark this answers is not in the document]
senarvi: Fixed.

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,553 @@
...
+        image_space_loss: bool = False,
+        overlap_loss_multiplier: float = 1.0,
+        class_loss_multiplier: float = 1.0,
+        confidence_loss_multiplier: float = 1.0
+    ):
```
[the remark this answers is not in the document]
senarvi: I fixed the type hints of all the constructors.

pl_bolts/models/detection/yolo/yolo_layers.py
akihironitta: > number of targets per image, which is typically not very large.

Sounds reasonable! Thanks for the update!

pl_bolts/datamodules/vocdetection_datamodule.py
```diff
@@ -152,16 +152,20 @@ def prepare_data(self) -> None:
         VOCDetection(self.data_dir, year=self.year, image_set="val", download=True)
 
     def train_dataloader(
-        self, batch_size: int = 1, image_transforms: Union[List[Callable], Callable] = None
+        self,
+        batch_size: int = 1,
+        transforms: List[Callable] = [],
```
kaushikb11, line 157: It's not considered good practice to have an empty list as default.

senarvi: That's true. For some reason I wanted to minimize changes outside the YOLO class, but you're right. I'll change the default value to None.

tests/models/test_detection.py
```diff
@@ -34,3 +38,122 @@ def test_fasterrcnn_bbone_train(tmpdir):
 
     trainer = pl.Trainer(fast_dev_run=True, default_root_dir=tmpdir)
     trainer.fit(model, train_dl, valid_dl)
+
+
+def _create_yolo_config_file(config_path):
+    config_file = open(config_path, 'w')
+    config_file.write(
+        '''[net]
```
Borda, line 46: can we rather have it in `tests/data` no need to be as a code

Borda (approved):
overall it looks good to me, mind check minor comments from @akihironitta

kaushikb11:
@senarvi The tests failing here are not related to these PR changes. It will be fixed in #580

senarvi:
> overall it looks good to me, mind check minor comments from @akihironitta

I think the comments from @akihironitta have been resolved. Let me know if there's still something.

tests/models/test_detection.py
senarvi: That's a good idea. I moved the test config to `tests/data`.

pl_bolts/models/detection/yolo/yolo_module.py
```diff
@@ -0,0 +1,551 @@
...
+
+        Args:
+            weight_file: A file object containing model weights in the Darknet binary format.
+        """
+        version = np.fromfile(weight_file, count=3, dtype=np.int32)
+        images_seen = np.fromfile(weight_file, count=1, dtype=np.int64)
+        print(
+            'Loading weights from Darknet model version {}.{}.{} that has been trained on {} '
+            'images.'.format(version[0], version[1], version[2], images_seen[0])
+        )
```
kaushikb11, line 276: @senarvi Is the `load_darnet_weights` functionality complete?

senarvi: It should be complete. I'm able to detect boxes using Darknet weights, and I'm able to continue training after loading Darknet weights. However, I got lower scores than when doing transfer learning using the Darknet implementation. I haven't got a chance to properly test it yet though, so I'm interested if you see any problems or if you're able to try it.

senarvi: I experimented with loading weights from a Darknet model. Training was very unstable, at least when using one of the truncated weight files. The Darknet weight files that are meant for transfer learning (such as [yolov4.conv.137](https://github.com/AlexeyAB/darknet/releases/download/darknet_yolo_v3_optimal/yolov4.conv.137)) are truncated up to the first detection layer, so that the number of object classes can be changed. The rest of the layers are initialized randomly. You can use them, but if you don't change the number of classes, you can load weights to all layers (such as [yolov4.weights](https://github.com/AlexeyAB/darknet/releases/download/darknet_yolo_v3_optimal/yolov4.weights)). I found that to work better. Otherwise the only way to make training stable was to use gradient clipping. It's strange because I think the Darknet implementation is not using gradient clipping. I spent some time trying to stabilize the operations such as divisions in IoU calculation, but to no avail.

senarvi:
I changed still one detail. The boxes are directly predicted in coordinates to the input image, not in the *[0, 1]* range. There are three reasons:
* a bit fewer scaling operations,
* consistency with torchvision models, and
* better resolution for representing coordinates with half-precision floats.

senarvi:
One more small improvement: image size from the configuration file was passed to `DetectionLayer` constructor, and any images fed to the network had to have the same size. Now the image size from the input batch is passed to its `forward()` method, so one can use a different size during inference, or - at least in theory - train with different image sizes.

jlevy44:
Hi there! This looks like an amazing PR. What is the timeline for merging?

kaushikb11:
The PR is good to go! But there are failing tests.
@akihironitta Could you please assist @senarvi with the failing tests, so we could go ahead with merging?

akihironitta:
@kaushikb11 Sure, I'll look into this within 24 hours :)

jlevy44:
Great, thanks @kaushikb11 @senarvi @akihironitta !

senarvi:
A test was failing because `/` was used on integer tensors, which is deprecated. I fixed it, but now I don't get a proper error message anymore, why the tests are stuck. Maybe just try rerunning them? The tests were already passing earlier, so there shouldn't be anything major anymore.

akihironitta:
@senarvi I approved and reran the tests. https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks was blocking you from running tests in CI.

pl_bolts/models/detection/yolo/yolo_layers.py
```diff
@@ -0,0 +1,487 @@
...
+        Returns:
+            A dictionary of training losses.
+        """
+        batch_size, height, width, boxes_per_cell, _ = boxes.shape
+        device = boxes.device
+        assert batch_size == len(targets)
+
+        # A multiplier for scaling image coordinates to feature map coordinates
+        grid_size = torch.tensor([width, height], device=device)
+        image_to_grid = grid_size / image_size
```
akihironitta, line 322:
```suggestion
        image_to_grid = torch.true_divide(grid_size, image_size)
```
Could you use `torch.true_divide()` instead of `/` here, too?

akihironitta (changes requested):
@senarvi Would you mind applying the suggestion below to pass the tests?

pl_bolts/models/detection/yolo/yolo_layers.py
senarvi: Yes, that seems to be the problem. Fixed.

senarvi:
Still one check fails to run. @akihironitta could you check what's wrong?

akihironitta:
All of these tests are failing due to the unavailability of MNIST datasets (http://yann.lecun.com/exdb/mnist/index.html is 503 at this moment), so I believe the errors are related to none of the changes in this PR. Let me fix the issue in another PR.

edgarriba:
@Borda what was the solution in the end for this mnist dataset issue ?

kaushikb11 (approved):

akihironitta:
I created an issue for the MNIST problem. #632

akihironitta (approved):
LGTM!

The failing tests are unrelated to the changes in this PR. It's being fixed in #633.

jlevy44:
Looks like #633 was merged. I still see a few failures. Anything I can do to help?

akihironitta:
@senarvi Could you merge master?

senarvi:
@akihironitta I merged master and I can see that the MNIST fix is in my [source branch](https://github.com/groke-technologies/pytorch-lightning-bolts/blob/yolo/pl_bolts/datasets/mnist_dataset.py), but the tests still fail with HTTP Error 503.

senarvi:
Merged master again, but now test_data_monitor.py::test_base_log_interval_* tests seem to be failing.

senarvi:
> Can there be an example of using the COCO dataset instead of the Pascal VOC dataset in the data module?

Currently there's no COCO data module. Adding a COCO data module is trivial, but one of the principles of this project is to keep pull requests simple and not to include several things in one pull request. So I don't know. Should I include a COCO data module in this pull request?

hammacktony:
> > Can there be an example of using the COCO dataset instead of the Pascal VOC dataset in the data module?
> 
> Currently there's no COCO data module. Adding a COCO data module is trivial, but one of the principles of this project is to keep pull requests simple and not to include several things in one pull request. So I don't know. Should I include a COCO data module in this pull request?

It might be beneficial to remove the data module part - since it is so tied to VOC, and have another PR or set of PRs that implement the different types of datamodules.

I tried to use the VOC one as a template for COCO, and I just couldn't get where I needed to. That's why I was asking.

senarvi:
> > > Can there be an example of using the COCO dataset instead of the Pascal VOC dataset in the data module?
> > 
> > 
> > Currently there's no COCO data module. Adding a COCO data module is trivial, but one of the principles of this project is to keep pull requests simple and not to include several things in one pull request. So I don't know. Should I include a COCO data module in this pull request?
> 
> It might be beneficial to remove the data module part - since it is so tied to VOC, and have another PR or set of PRs that implement the different types of datamodules.
> 
> I tried to use the VOC one as a template for COCO, and I just couldn't get where I needed to. That's why I was asking.

What exactly do you suggest removing? There's a small change to the existing VOC data module that allows passing transforms for both images and target boxes, for example, resizing both the image and its target boxes. This lets us see that training on VOC data works. The dummy CLI is similar to that of the currently only object detection model, FasterRCNN, and is meant more like sample code rather than a useful tool. I agree that a pull request for COCO data module would be useful, and could also update the CLIs so that they support COCO data.

Implementing a data module for COCO should be quite easy using the VOC data module as a template. Just remember to convert the coordinates from (x1, y1, width, height) to (x1, y1, x2, y2).

Borda:
@senarvi could you check the failing tests? and mind allow edit for the maintainers... :rabbit:

senarvi:
> @senarvi could you check the failing tests? and mind allow edit for the maintainers... 🐰

@Borda the test was failing because I was introduced a bug in the MSE loss, when I fixed the IoU loss. Both should work now. How do I allow edit for the maintainers?

senarvi:
> > @senarvi could you check the failing tests? and mind allow edit for the maintainers... 🐰
> 
> @Borda the test was failing because I introduced a bug in the MSE loss, when I fixed a problem in the IoU loss. Both should work now. How do I allow edit for the maintainers?

It seems that "Allow edits from maintainers" is not possible when the source repository is under an organization.

senarvi:
@Borda I noticed also the formatting issues and fixed them by running the new formatting tools.

senarvi:
`VOCDetectionDataModule` was updated to better conform to the new data module interface. The constructor takes batch size and the transforms take image and target. We can now pass the data module to Trainer, instead of the dataloaders. `train_dataloader()` and `val_dataloader()` still support image-only transforms, but I don't see any need for using them.

senarvi:
The test suite found still one problem with integer division with an older PyTorch version. @Borda could you restart the CI? I don't see the error when running the tests with my PyTorch installation.

akihironitta:
@senarvi Hi, thanks again for all your work and patience! Just restarted CI again. I'll try to be active on this PR for the next few days until we get this PR merged :]

senarvi:
@akihironitta correct me if I'm wrong, but the remaining error was something about CIFAR dataset and not related to this pull request.

CHANGELOG.md
```diff
@@ -88,6 +88,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added Pix2Pix model ([#533](https://github.com/PyTorchLightning/lightning-bolts/pull/533))
+- Added YOLO model ([#552](https://github.com/PyTorchLightning/pytorch-lightning-bolts/pull/552))
```
akihironitta, line 91: Can you move this line to the unreleased section?

pl_bolts/models/detection/__init__.py
```diff
@@ -1,7 +1,6 @@
 from pl_bolts.models.detection import components
 from pl_bolts.models.detection.faster_rcnn import FasterRCNN
+from pl_bolts.models.detection.yolo.yolo_config import YOLOConfiguration
+from pl_bolts.models.detection.yolo.yolo_module import YOLO
 
-__all__ = [
-    "components",
-    "FasterRCNN",
-]
+__all__ = ["components", "FasterRCNN", "YOLOConfiguration", "YOLO"]
```
akihironitta, line 6: nit-picking :smile: 
```suggestion
__all__ = [
    "components",
    "FasterRCNN",
    "YOLOConfiguration",
    "YOLO",
]
```

akihironitta (commented):
> @akihironitta correct me if I'm wrong, but the remaining error was something about CIFAR dataset and not related to this pull request.

@senarvi You're right that all of the tests here have passed. Once the following comments gets resolved, I believe we can land this PR finally!

senarvi:
@akihironitta ok, done.

PR #655, opened 2021-05-30
Fix - Avoid two softmax calls in LogisticRegression

### LINKED ISSUE #654: Softmax applied twice when computing loss in LogisticRegression
## 🐛 Bug

In the `LogisticRegression` model, loss is computed as the negative log-likelihood of the log-softmax of the softmax of the linear classifier outputs. I.e. `loss = F.nll_loss(F.log_softmax(F.softmax(self.linear(x))), y)`

#### Code sample

```Python
model = LogisticRegression(42, 42)
batch = (torch.rand(42), torch.rand(42))
model.training_step(batch, 0)
```

### Expected behavior

Network loss should be computed as the cross entropy of predictions and targets (negative log likelihood of the log-softmax of the linear classifier outputs). I.e. `loss = F.nll_loss(F.log_softmax(self.linear(x)), y)`

### Environment

 - PyTorch Version (e.g., 1.0): `1.8.1`
 - OS (e.g., Linux): `RHEL 8.4`
 - How you installed PyTorch (`conda`, `pip`, source): `pip`
 - Build command you used (if compiling from source): `N/A`
 - Python version: `3.9.2`
 - CUDA/cuDNN version: `N/A`
 - GPU models and configuration: `N/A`
 - Any other relevant information: `None`

### Additional context

This bug should not cause any serious issue, however it deviates from expected behaviour and reduces value separation which has potential to negatively impact training.
The PyTorch implementation of Cross Entropy Loss includes the log-softmax, as such it expects raw unnormalized scores for each class ([docs](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html), [functional](https://pytorch.org/docs/stable/nn.functional.html#cross-entropy))
Happy to submit a PR to resolve if deemed appropriate

pierrenodet:
Avoid two softmax calls in LogisticRegression

## What does this PR do?

Fixes #654

### DIFF
--- a/pl_bolts/models/regression/logistic_regression.py
+++ b/pl_bolts/models/regression/logistic_regression.py
@@ -53,7 +53,7 @@ class LogisticRegression(pl.LightningModule):
         # flatten any input
         x = x.view(x.size(0), -1)
 
-        y_hat = self(x)
+        y_hat = self.linear(x)
 
         # PyTorch cross_entropy function combines log_softmax and nll_loss in single function
         loss = F.cross_entropy(y_hat, y, reduction='sum')


Borda (approved):

tchaton (approved):
LGTM !

garryod:
Hello @pierrenodet, could we make the same changes in `validation_step()` and `test_step()` such that the losses are the same? (Will then need to add a 'F.softmax()' inside the 'accuracy()' call as it expects probabilities)

PR #663, opened 2021-06-16
Fixed double softmax bug in SSLEvaluator

### LINKED ISSUE #662: Softmax applied twice in SSLOnlineEvaluator callback
## 🐛 Bug

It looks like the `SSLEvaluator` module includes a `nn.Softmax()` operation as its last layer, but the `SSLOnlineEvaluator` callback, which trains an `SSLEvaluator` module, uses `F.cross_entropy_loss` as the loss function. This is mathematically incorrect and could lead to systematically worse prediction performance, since `F.cross_entropy_loss` expects raw, unnormalized scores instead of probabilities.

It seems like this could be fixed by simply removing the `nn.Softmax()` module from `SSLEvaluator`, as long as nothing else depends on that being there, or (somewhat less ideally) just taking the log of the probabilities from `SSLEvaluator` and then using `F.nll_loss`.

UPDATE: The bug also seems to appear in `SSLFinetuner`. It looks like removing the `nn.Softmax()` module from `SSLEvaluator` should fix the bug, although we'll have to make sure that the logits get softmax applied to them before being fed into any metrics.

### To Reproduce

Look at [evaluator.py](https://github.com/PyTorchLightning/lightning-bolts/blob/master/pl_bolts/models/self_supervised/evaluator.py) and then look at [ssl_online.py](https://github.com/PyTorchLightning/lightning-bolts/blob/master/pl_bolts/callbacks/ssl_online.py).

norabelrose:
## What does this PR do?

Fixes #662

### DIFF
--- a/pl_bolts/callbacks/ssl_online.py
+++ b/pl_bolts/callbacks/ssl_online.py
@@ -102,8 +102,8 @@ class SSLOnlineEvaluator(Callback):  # pragma: no cover
         representations = representations.detach()
 
         # forward pass
-        mlp_preds = pl_module.non_linear_evaluator(representations)  # type: ignore[operator]
-        mlp_loss = F.cross_entropy(mlp_preds, y)
+        mlp_logits = pl_module.non_linear_evaluator(representations)  # type: ignore[operator]
+        mlp_loss = F.cross_entropy(mlp_logits, y)
 
         # update finetune weights
         mlp_loss.backward()
@@ -111,7 +111,7 @@ class SSLOnlineEvaluator(Callback):  # pragma: no cover
         self.optimizer.zero_grad()
 
         # log metrics
-        train_acc = accuracy(mlp_preds, y)
+        train_acc = accuracy(mlp_logits.softmax(-1), y)
         pl_module.log('online_train_acc', train_acc, on_step=True, on_epoch=False)
         pl_module.log('online_train_loss', mlp_loss, on_step=True, on_epoch=False)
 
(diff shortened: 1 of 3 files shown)


pl_bolts/models/self_supervised/evaluator.py
```diff
@@ -10,23 +10,12 @@ def __init__(self, n_input, n_classes, n_hidden=512, p=0.1):
         self.n_hidden = n_hidden
         if n_hidden is None:
             # use linear classifier
-            self.block_forward = nn.Sequential(
-                Flatten(),
-                nn.Dropout(p=p),
-                nn.Linear(n_input, n_classes, bias=True),
-                nn.Softmax(),
-            )
+            self.block_forward = nn.Sequential(Flatten(), nn.Dropout(p=p), nn.Linear(n_input, n_classes, bias=True))
```
Borda, line 13:
```suggestion
            self.block_forward = nn.Sequential(Flatten(), nn.Dropout(p=p), nn.Linear(n_input, n_classes, bias=True),)
```

pl_bolts/models/self_supervised/evaluator.py
```diff
@@ -10,23 +10,12 @@ def __init__(self, n_input, n_classes, n_hidden=512, p=0.1):
...
-                Flatten(),
-                nn.Dropout(p=p),
-                nn.Linear(n_input, n_hidden, bias=False),
-                nn.BatchNorm1d(n_hidden),
-                nn.ReLU(inplace=True),
-                nn.Dropout(p=p),
-                nn.Linear(n_hidden, n_classes, bias=True),
-                nn.Softmax(),
+                Flatten(), nn.Dropout(p=p), nn.Linear(n_input, n_hidden, bias=False), nn.BatchNorm1d(n_hidden),
+                nn.ReLU(inplace=True), nn.Dropout(p=p), nn.Linear(n_hidden, n_classes, bias=True)
```
Borda, line 18:
```suggestion
                nn.ReLU(inplace=True), nn.Dropout(p=p), nn.Linear(n_hidden, n_classes, bias=True),
```

Borda (approved):

ethanwharris (approved):
LGTM :smiley:

pl_bolts/callbacks/ssl_online.py
```diff
@@ -102,16 +102,16 @@ def on_train_batch_end(
         representations = representations.detach()
 
         # forward pass
-        mlp_preds = pl_module.non_linear_evaluator(representations)  # type: ignore[operator]
-        mlp_loss = F.cross_entropy(mlp_preds, y)
+        mlp_logits = pl_module.non_linear_evaluator(representations)  # type: ignore[operator]
```
awaelchli, line 105: If we look at the [SSLEvaluator](https://github.com/PyTorchLightning/lightning-bolts/blob/2ba3f62709b50604c29448f4ad620449796bce06/pl_bolts/models/self_supervised/evaluator.py) we find that it does not return logits.

The name should reflect what the output is :)

Borda:
> LGTM
> ignore my previous comment

I do not see any :P

PR #728, opened 2021-09-06
Improve SSLOnlineEvaluator

zhengyu-yang:
## What does this PR do?

Make `SSLOnlineEvaluator` supports DDP and checkpoint.

1. Current `SSLOnlineEvaluator` does not support DDP. Weights of classification layers are not synced across different processes.
2. Current `SSLOnlineEvaluator` will break `resume_from_checkpoint` functionality of `Trainer` because `SSLOnlineEvaluator` add an additional `non_linear_evaluator` to the main `pl_module` during `on_pretrain_routine_start`. This will cause the statedict to mismatch when `resume_from_checkpoint`.
3. Both `num_classes` and `dataset` arguments can be inferred during runtime. We do not need to pass them in during initialization.
4. When train the online evaluator, the main `pl_module` should use `eval` mode.

This pull request deals with the above mentioned issues. 

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/lightning-bolts/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you write any new necessary tests? \[not needed for typos/docs\]
- [ ] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/lightning-bolts/blob/master/CHANGELOG.md)?

## PR review


zhengyu-yang:
Need to add `convert_sync_batchnorm` in case a non-linear online evaluator is used.

pl_bolts/callbacks/ssl_online.py
```diff
@@ -1,74 +1,96 @@
...
-        self.optimizer: Optimizer
 
-        self.z_dim = z_dim
-        self.num_classes = num_classes
-        self.dataset = dataset
+        self.optimizer: Optional[Optimizer] = None
+        self.online_evaluator: Optional[SSLEvaluator] = None
+        self.z_dim: Optional[int] = None
+        self.num_classes: Optional[int] = None
+        self.dataset: Optional[str] = None
```
Borda, line 56: how these are set?

pl_bolts/callbacks/ssl_online.py
```diff
@@ -1,74 +1,96 @@
...
         self.hidden_dim = hidden_dim
         self.drop_p = drop_p
-        self.optimizer: Optimizer
 
-        self.z_dim = z_dim
-        self.num_classes = num_classes
-        self.dataset = dataset
+        self.optimizer: Optional[Optimizer] = None
+        self.online_evaluator: Optional[SSLEvaluator] = None
+        self.z_dim: Optional[int] = None
```
Borda, line 54: seems overwrite line 48

pl_bolts/callbacks/ssl_online.py
zhengyu-yang: They are set during `setup()`

pl_bolts/callbacks/ssl_online.py
zhengyu-yang: I forgot to remove line 54.

pl_bolts/callbacks/ssl_online.py
zhengyu-yang: See [here](https://github.com/zhengyu-yang/lightning-bolts/blob/edbfca7ee9a283df2d9aee3e8705cb1c53bffe3f/pl_bolts/callbacks/ssl_online.py#L59-L61)

Borda: well, can we still keep the init arguments and if needed set default as None?
or pls what is the reason for removing these arguments from init?

zhengyu-yang: This is more of a design choice. As this is adapted from my research code, and in my experiments, I try to remove any redundancy in configuration. 
I can add them back for backward compatibility.

Borda: > I can add them back for backward compatibility. Do you think it is necessary?

yes, pls add it back as it does not hurt you and we do not break any user code...

SeanNaren: Thanks @zhengyu-yang as @Borda suggested, it would be nice to keep the signature the same for BW compat :)

SeanNaren (commented):
Great work @zhengyu-yang! Everything else looks good to me, however am curious if the online evaluator works well considering the entire module is wrapped in DDP, and then the online evaluator is also wrapped.

Borda:
@zhengyu-yang @ananyahjha93 any progress here? :rabbit:

zhengyu-yang:
I have made `num_classes ` and `dataset ` optional

Borda:
@zhengyu-yang @SeanNaren any update here? :rabbit:

zhengyu-yang:
> @zhengyu-yang @SeanNaren any update here? 🐰
Is there any todo on my side?

Borda:
> Is there any todo on my side?

seems that the GPU tests are failing...

ant0nsc:
@zhengyu-yang thanks very much for this super helpful change, we need that right now, too!
Just one question around how you designed that: Why are you not loading the state dict of the model/optimizer directly in `on_load_checkpoint`? Rather, you are saving it in a private member, and only use it in `on_pretrain_routine_start`. Why is that?

Issue #839, opened 2022-07-21
Major revision of Bolts


otaj:
Let's revision Bolts and breathe some fresh air into them! As outlined in #819 and on a Slack channel, we will revisit every single feature within Bolts.

Please sign up for a feature which you'd like to tackle. Once you do so, I will attach your name in the list and you will be expected to open a PR within two weeks. It might be useful to tackle multiple things at once as the feature list is everything top-level from the repository.

## Criteria

The criteria for acceptance are simple:
1. Feature has to be properly tested. Unit tests are essential, integration tests are also good, but we understand there might be situations when they are impossible.
2. Feature has to be compatible with the latest set of Lightning libraries (lightning, flash, torchmetrics). This fact is also tested. This also means outputting things in formats consumable by other projects, namely Flash.
3. Feature cannot be duplicated across our codebase, codebase of other Lightning projects and other more mature OSS projects. For example, there are currently five functions/classes that implement conv3x3 in this repository.
4. And of course, once you're done, remove the decorator `@under_review` from the selected feature :tada: 

## List of features to be reviewed

<details>
<summary>Features</summary>

- [ ] `pl_bolts.callbacks.byol_updates.BYOLMAWeightUpdate`
- [ ] `pl_bolts.callbacks.data_monitor.DataMonitorBase`
- [ ] `pl_bolts.callbacks.data_monitor.ModuleDataMonitor`
- [ ] `pl_bolts.callbacks.data_monitor.shape2str`
- [ ] `pl_bolts.callbacks.data_monitor.TrainingDataMonitor`
- [ ] `pl_bolts.callbacks.knn_online.concat_all_gather`
- [ ] `pl_bolts.callbacks.knn_online.KNNOnlineEvaluator`
- [ ] `pl_bolts.callbacks.printing.dicts_to_table`
- [ ] `pl_bolts.callbacks.printing.PrintTableMetricsCallback`
- [ ] `pl_bolts.callbacks.sparseml.SparseMLCallback`
- [ ] `pl_bolts.callbacks.ssl_online.set_training`
- [ ] `pl_bolts.callbacks.ssl_online.SSLOnlineEvaluator`
- [ ] `pl_bolts.callbacks.torch_ort.ORTCallback`
- [ ] `pl_bolts.callbacks.variational.LatentDimInterpolator`
- [ ] `pl_bolts.callbacks.verification.base.VerificationBase`
[...]

Issue #904, opened 2022-10-11
Central location for transforms


matsumotosan:
Data transform classes for various models are included in their own folders in `pl_bolts/models`. It may be better to have them in `pl_bolts/transforms` instead.

Issue discussed with @otaj

senarvi:
I was going to suggest the same thing. Also, at least SimCLR and MoCo can use the same transforms. Maybe we want to provide some transforms that can be used to replicate the paper results on some dataset, but do we want to provide a separate set of transforms for every model and every dataset?

matsumotosan:
In general, I think the existing transforms are based on what the original papers specified (for whatever datasets they used). If a model by default uses transforms from another (i.e. BYOL using SimCLR transforms), I think we can just import SimCLRTransforms for BYOL instead of creating a copy and calling it BYOLTransforms. I think creating transforms for every model for every dataset might be too much though.

I'm all for standardizing how transforms are set up. I'm almost done moving everything over to `pl_bolts/transforms` so I'll let you know when I'm finished with that.

otaj:
Hi, @matsumotosan, that would be great, thank you for kicking it off!

matsumotosan:
All self-supervised transforms have been moved #905

matsumotosan:
Commenting so this issue doesn't get closed.

All changes have been made in #905. Waiting for CI to be fixed so we can finish testing.

PR #928, opened 2022-10-31
Revision of the MoCo SSL model

senarvi:
## What does this PR do?

Related to issue #839 that discusses a major revision of Bolts, this pull request refactors the Momentum Contrast (MoCo) SSL model. That issue will probably become impossible to follow if all the details are discussed there, so I'll write my suggestions below. I would be happy to hear some feedback.

## Before submitting

- [x] Was this **discussed/approved** via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/lightning-bolts/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure your PR does **only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes?
- [x] Did you write any **new necessary tests**? \[not needed for typos/docs\]
- [x] Did you verify new and **existing tests** pass locally with your changes?
- [ ] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/lightning-bolts/blob/master/CHANGELOG.md)?

## PR review

- [x] Is this pull request **ready for review**? (if not, please submit in draft mode)

## Did you have fun?


senarvi:
@Atharva-Phatak SimCLR is a very similar method, so it makes sense to coordinate our efforts, and maybe other SSL methods too, so that we could make the interfaces as consistent as possible. We could also place both models in `pl_bolts.models.ssl.contrastive` module in order to be able to share code, if that's a good idea.

I think it would be important to make the models generic so that one can train
* any network, instead of selecting one of a set of options
* using any projection head
* using any optimizer and learning rate schedule
* using any data loader and transforms

Issues #904 already suggests moving transforms to a central location. I don't know if we need separate transforms for each model. More or less identical transforms are used with both methods, and the user doesn't necessarily want to stick to the default transforms anyway.

We should fix the format that the data loader should produce for contrastive learning. The format could be as similar as possible to the format that we use for classification and object detection. At least for object detection we use a tuple `(images, targets)`, where `images` is a list of image tensors and SSL models can simply ignore `targets`. In this pull request `images` is expected to be a nested list `[[image1A, image1B], [image2A, image2B], [image3A, image3B], ...]`.

I added an example CLI application using LightningCLI, but IMO there's no point in trying to create a comprehensive training tool in each module.

I renamed the class to MoCo. I think v2 just adds a projection head, so the same class supports both versions.

senarvi:
The old way of detecting whether DDP is in use didn't work, since the DDP strategy class is now `DDPStrategy`, not `DDPPlugin`. Using `isinstance()` is pretty error-prone. I wonder if we could use `if torch.distributed.is_available() and torch.distributed.is_initialized()`, or is it important to detect DDP as opposed to other distributed training strategies?

Atharva-Phatak:
@senarvi Perfect, that is what my intention is essentially we add Projection heads and Losses required by SSL models. Backbones can be implemented by the user if they are custom or else imported from timm or something similar. 

All in all here is what I suggest

1) Single place for SSL projection heads
2) Single place for SSL losses
3) Single place for transforms as well.
3) General PL module for training any custom SSL model :)

otaj:
Hi, @senarvi, @Atharva-Phatak. I fully support @Atharva-Phatak's suggestion, if you can coordinate on that, that would be great!

Btw, @senarvi, since your fork is on an organization account (`groke-technologies`) and not on personal account, noone except people belonging to that organization have write access to it. While that makes sense, I'd say it makes it hard for us maintainers to help you with those PRs since we cannot merge master.

senarvi:
@otaj ok that explains it. But it's fine, I'll just keep merging master myself.

ArnolFokam:
> @senarvi Perfect, that is what my intention is essentially we add Projection heads and Losses required by SSL models. Backbones can be implemented by the user if they are custom or else imported from timm or something similar.
> 
> All in all here is what I suggest
> 
> 1. Single place for SSL projection heads
> 2. Single place for SSL losses
> 3. Single place for transforms as well.
> 4. General PL module for training any custom SSL model :)

Happy to help with this as well. Something we might want to be cautious of (if we follow this route) is the compatibility between the transforms and losses of various SSL methods.

Borda:
@senarvi seems I can't push changes to this branch/PR, could you pls merge master here? 🐿️

Borda:
@senarvi how is it going here? :rabbit:

Borda:
FYI: transforms were moved to on sub-package

senarvi:
@Borda thanks for reminding me. I noticed that the transforms were moved to a different location, but there seems to be something else. Maybe a circular import. I'm looking into it.

Borda:
> Maybe a circular import. I'm looking into it.

Could be, mind allowing maintainers to push to your PR so can try to help? :)

senarvi:
@Borda the problem is, this comes from an organization, so I don't know if it's possible to allow developers to push to the branch. Now I hope that I was able to merge it. Fingers crossed that the tests will pass.

Borda:
One test failing, looks promising: FAILED `tests/models/self_supervised/test_models.py::test_moco - ValueError: Contrastive training expects a tuple or a list of transformed images, got Tensor.`

Borda (approved):

senarvi:
> One test failing, looks promising: FAILED `tests/models/self_supervised/test_models.py::test_moco - ValueError: Contrastive training expects a tuple or a list of transformed images, got Tensor.`

Let me look into that tomorrow. It means that the test is not providing data in the correct format. Unfortunately I wasn't able to run the tests locally. I'll have to find out how to do that.

senarvi:
@Borda the problem was that the data module used in the unit tests provided 4-dimensional tensors of image pairs, while the default ImageNet data module provided tuples of 3-dimensional tensors and that's what the input validator expected too. Now it supports both.

There was also a new unit test that checks the CLI scripts. It failed because ImageNet cannot be downloaded automatically, so I changed the CLI application to use CIFAR10 instead.

If I'm correct, now all the unit tests pass. I can only see some Docker errors now. But let me know if there's still something that needs fixing.

Borda:
@senarvi the GPU start issue is fixed on master...

senarvi:
@Borda merged master and now everything passes.

src/pl_bolts/models/detection/retinanet/retinanet_module.py
```diff
@@ -136,13 +136,17 @@ def configure_optimizers(self):
 @under_review()
 def cli_main():
     try:  # Backward compatibility for Lightning CLI
-        from pytorch_lightning.cli import LightningCLI  # PL v1.9+
-    except ImportError:
-        from pytorch_lightning.utilities.cli import LightningCLI  # PL v1.8
+        import pytorch_lightning.cli
+
+        cli_class: Any = getattr(pytorch_lightning.cli, "LightningCLI")  # PL v1.9+
+    except Exception:
```
Borda, line 142: can we be more specific here with exception?

src/pl_bolts/models/detection/retinanet/retinanet_module.py
```diff
@@ -136,13 +136,17 @@ def configure_optimizers(self):
...
+        import pytorch_lightning.utilities.cli
+
+        cli_class = getattr(pytorch_lightning.utilities.cli, "LightningCLI")  # PL v1.8
```
Borda, line 145:
```suggestion
        from pytorch_lightning.cli import LightningCLI  # PL v1.9+
    except ImportError:
        from pytorch_lightning.utilities.cli import LightningCLI  # PL v1.8
```

src/pl_bolts/models/detection/retinanet/retinanet_module.py
```diff
@@ -136,13 +136,17 @@ def configure_optimizers(self):
...
 
     from pl_bolts.datamodules import VOCDetectionDataModule
 
-    LightningCLI(RetinaNet, VOCDetectionDataModule, seed_everything_default=42)
+    cli_class(RetinaNet, VOCDetectionDataModule, seed_everything_default=42)
```
Borda, line 149:
```suggestion
    LightningCLI(RetinaNet, VOCDetectionDataModule, seed_everything_default=42)
```

src/pl_bolts/models/detection/yolo/yolo_module.py
```diff
@@ -614,4 +613,13 @@ def _resize(self, image: Tensor, target: TARGET) -> Tuple[Tensor, TARGET]:
 
 
 if __name__ == "__main__":
-    LightningCLI(CLIYOLO, ResizedVOCDetectionDataModule, seed_everything_default=42)
+    try:  # Backward compatibility for Lightning CLI
```
Borda, line 616: The same here...

senarvi:
@Borda the thing that you commented on is a real PITA to get working so that mypy won't complain in any case. There's a lot of talk about this on GitHub, see for example [this](https://github.com/python/mypy/issues/8823). The problem is that mypy doesn't consider the `try..except` and complains about LightningCLI being already defined. If `# type: ignore` comments are added, it will complain about an unused 'type: ignore' comment, because always one of the comments is unnecessary. I'm not sure if there's a fix in the latest mypy. At least with mypy 1.1.1 the code that you suggested won't pass.

senarvi:
@Borda also, the exception can be either an `ImportError` or an `AttributeError`, so it's an ugly workaround indeed.

senarvi:
@Borda should we go like this? I'm fine with anything that makes mypy happy.

Borda:
> I'm fine with anything that makes mypy happy.

lets mypy/typing address in separate PR :)

src/pl_bolts/models/detection/retinanet/retinanet_module.py
Borda: @senarvi lets use standard import pattern

senarvi:
@Borda fine. Actually it was necessary to only support the current location (pytorch_lightning.cli) to make the tests pass. Now it's clean.

Borda (approved):

